PR #1, opened 2023-08-15
Add Transcript trait, with PoseidonTranscript impl

arnaucube:
Add Transcript trait, with PoseidonTranscript impl, add also the PoseidonTranscriptVar (gadget).


src/lib.rs
```diff
@@ -1,11 +1,20 @@
...
+
 #[derive(Debug)]
 pub enum Error {}
 
-pub trait Transcript<C: CurveGroup>: Debug {
-    // todos
+pub trait Transcript<F: PrimeField> {
+    type TranscriptConfig: Debug;
+
+    fn new(config: &Self::TranscriptConfig) -> Self;
```
arnaucube, line 13: here I was not sure if to use `new` or `setup`

han0110: I think `new` should be good

src/transcript.rs
```diff
@@ -0,0 +1,121 @@
+use ark_ff::PrimeField;
```
han0110, line 1: perhaps remove empty lines between import?

src/transcript.rs
```diff
@@ -0,0 +1,121 @@
...
+        Ok(c[0].clone())
+    }
+    pub fn get_challenges(&mut self, n: usize) -> Result<Vec<FpVar<F>>, SynthesisError> {
+        let c = self.sponge.squeeze_field_elements(n)?;
+        self.sponge.absorb(&c)?;
+        Ok(c)
+    }
+}
+
+/// WARNING the method poseidon_test_config is for tests only
```
han0110, line 69: If it's only used in tests, perhaps move it into mod `tests`?

han0110 (approved):
LGTM! Just some nitpicks

src/transcript.rs
arnaucube: Right! moved into mod `tests`

src/transcript.rs
arnaucube: Right, updated

PR #3, opened 2023-08-17
Feature/traits

arnaucube:


src/transcript.rs
```diff
@@ -0,0 +1,121 @@
+use ark_crypto_primitives::sponge::{
```
arnaucube, line 1: I was thinking about maybe this file could be `poseidon_transcript.rs`, since just implements the `PoseidonTranscript` for the trait `Transcript` defined in `lib.rs`.
Because we might also implement a `keccak_transcript.rs` in the future, as for example in [microsoft/Nova](https://github.com/microsoft/Nova) it is used only the keccak transcript since is not needed to be computed in-circuit.
How do you see it to rename this file to `poseidon_transcript.rs`?

arnaucube: The approach that you mention makes more sense and is cleaner, updated it in https://github.com/privacy-scaling-explorations/folding-schemes/pull/3/commits/e9fd878170f47c87bc0887cdb43bd05c61086a88 with that mod structure.

PR #4, opened 2023-08-17
Port CI from `halo2curves`

han0110:
That was originally introduced by @ huitseeker

Also test in my repo and the CI passes with current codebase, see https://github.com/han0110/folding-schemes/actions/runs/5889212647/job/15971929113


arnaucube (approved):
Great! ^^

PR #6, opened 2023-08-22
Add initial CCS mod

arnaucube:
Add initial CCS mod:
- port initial CCS structure with methods from multifolding-poc
- add R1CS helper methods, which will be used in Nova impl


han0110 (approved):
LGTM!

PR #7, opened 2023-08-23
Implement Nova's NIFS.P & NIFS.V

arnaucube:
Since R1CS is almost the same than CCS, and Nova is described using R1CS, Nova impl uses R1CS for the internal Nova code, but it has almost no cost to go from the higher level CCS instance into the internal R1CS representation of the same instance, so to the world outside the mod all can be with CCS.

This does not implement the [FoldingScheme](https://github.com/privacy-scaling-explorations/folding-schemes/blob/main/src/lib.rs#L27) trait for Nova, just the NIFS logic which will be used in the future by the FoldingScheme trait impl in Nova.

Note regarding the directories structure: initially I was starting `nova` & `hypernova` inside the `ccs` directory, but then I thought to better have `ccs` structs & helpers in the `ccs` dir, but the `nova` & `hypernova` code under a `folding` directory. So for example we can have in the `folding` dir the implementation of `protogalaxy` that can use both `ccs` and `plonkish`.

### DIFF
--- a/src/folding/nova/mod.rs
+++ b/src/folding/nova/mod.rs
@@ -1,0 +1,62 @@
+use ark_crypto_primitives::sponge::Absorb;
+use ark_ec::{CurveGroup, Group};
+use ark_std::fmt::Debug;
+use ark_std::{One, Zero};
+
+use crate::pedersen::{Params as PedersenParams, Pedersen};
+
+pub mod nifs;
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct CommittedInstance<C: CurveGroup> {
+    pub cmE: C,
+    pub u: C::ScalarField,
+    pub cmW: C,
+    pub x: Vec<C::ScalarField>,
+}
+impl<C: CurveGroup> CommittedInstance<C> {
+    pub fn empty() -> Self {
+        CommittedInstance {
+            cmE: C::zero(),
+            u: C::ScalarField::one(),
+            cmW: C::zero(),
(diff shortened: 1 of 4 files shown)


src/folding/nova/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+pub struct CommittedInstance<C: CurveGroup> {
+    pub cmE: C,
+    pub u: C::ScalarField,
+    pub cmW: C,
+    pub x: Vec<C::ScalarField>,
+}
+impl<C: CurveGroup> CommittedInstance<C> {
+    pub fn empty() -> Self {
+        CommittedInstance {
+            cmE: C::generator(),
```
han0110, line 20: `C::zero` seems closer to `empty`

src/folding/nova/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+            u: C::ScalarField::one(),
+            cmW: C::generator(),
```
han0110, line 22: ditto

han0110 (approved):
LGTM. The newly created `mod folding` also makes sense to me!

src/folding/nova/mod.rs
arnaucube: Absolutely! Thanks for pointing it out ^^, updated

src/folding/nova/mod.rs
arnaucube: Thanks! updated

PR #9, opened 2023-08-25
Implement CycleFold's EC RLC check circuit

arnaucube:
The idea is that `src/folding/circuits` contains the circuits & gadgets shared across the different folding schemes, and then each folding scheme will have a `circuits` module for their specific circuits, eg. `src/folding/nova/circuits.rs`, `src/folding/hypernova/circuits.rs`, etc.


han0110 (approved):
LGTM!

PR #28, opened 2023-09-21
Refactor Matrix vectors and Logical operation

ashWhiteHat:
Hi there
I refactored Pedersen struct and Matrix vectors operation.

- Pedersen

Pedersen struct doesn't have parameters now and every time it takes params as argument.
I refactored it to store params inside of struct.

- Matrix

There are some unnecessary initialization of struct.
I replaced it with one line array method.

- NIFT

There are some unnecessary condition branch.
I replaced it with simpler logical operations.

I would appreciate it if you could confirm.
Thank you.

### DIFF
--- a/src/ccs/mod.rs
+++ b/src/ccs/mod.rs
@@ -91,6 +91,7 @@ impl<C: CurveGroup> CCS<C> {
             M: vec![r1cs.A, r1cs.B, r1cs.C],
         }
     }
+
     pub fn to_r1cs(self) -> R1CS<C::ScalarField> {
         R1CS::<C::ScalarField> {
             l: self.l,
--- a/src/ccs/r1cs.rs
+++ b/src/ccs/r1cs.rs
@@ -53,6 +53,7 @@ pub struct RelaxedR1CS<F: PrimeField> {
     pub u: F,
     pub E: Vec<F>,
 }
+
 impl<F: PrimeField> RelaxedR1CS<F> {
     /// check that a RelaxedR1CS structure is satisfied by a z vector. Only for testing.
     pub fn check_relation(&self, z: &[F]) -> Result<(), Error> {
--- a/src/folding/nova/nifs.rs
+++ b/src/folding/nova/nifs.rs
@@ -136,16 +136,11 @@ where
         cmT: &C,
     ) -> Result<(), Error> {
         let r2 = r * r;
-        if ci3.cmE != (ci1.cmE + cmT.mul(r) + ci2.cmE.mul(r2)) {
-            return Err(Error::NotSatisfied);
-        }
-        if ci3.u != ci1.u + r * ci2.u {
-            return Err(Error::NotSatisfied);
-        }
-        if ci3.cmW != (ci1.cmW + ci2.cmW.mul(r)) {
-            return Err(Error::NotSatisfied);
-        }
-        if ci3.x != vec_add(&ci1.x, &vec_scalar_mul(&ci2.x, &r))? {
+        if ci3.cmE != (ci1.cmE + cmT.mul(r) + ci2.cmE.mul(r2))
+            || ci3.u != ci1.u + r * ci2.u
+            || ci3.cmW != (ci1.cmW + ci2.cmW.mul(r))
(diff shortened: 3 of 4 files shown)


src/pedersen.rs
```diff
@@ -14,35 +13,29 @@ pub struct Proof<C: CurveGroup> {
 }
 
 #[derive(Debug, Clone, Eq, PartialEq)]
-pub struct Params<C: CurveGroup> {
+pub struct Pedersen<C: CurveGroup> {
```
arnaucube, line 16: This comment references to all the changes from the commit [22343757b](https://github.com/privacy-scaling-explorations/folding-schemes/pull/28/commits/22343757b0c6a9f322750527e07b65c99b29c7b5).

Here the original code was following the arkworks style of passing the parameters, as example: arkwors/crypto-primitives: https://github.com/arkworks-rs/crypto-primitives/blob/1e4f16cddba8d18f4eb219fa33ac25b2c852f198/src/commitment/pedersen/mod.rs#L63

or as it is done also in microsoft/Nova (non-arkworks): https://github.com/microsoft/Nova/tree/200b47edab190c06a44cbd87a63d34a444184a41/src/provider/pedersen.rs#L196, where `CommitmentKey` is kind of the `Params` in our case.

On both cases of the previous links, the parameters are passed as parameter and there is no `Pedersen` structure holding them.

Do you have any motivation to change it into how it is done in this PR having the `pedersen` object instead as how it was done before this PR and how it is done in arkworks & microsoft/Nova?

arnaucube (commented):
Thanks for the contribution ^^

The oneliners for `utils/vec.rs` look very good! Also the conditionals grouped in the NIFS verifications make lot of sense, thanks!

I do have one question regarding the pedersen changes, left as a comment.

src/pedersen.rs
ashWhiteHat: Thank you for the review and references.

According to references, it depends on abstraction degree.

In arkworks-rs, they do abstraction with [CommitmentScheme](https://github.com/arkworks-rs/crypto-primitives/blob/1e4f16cddba8d18f4eb219fa33ac25b2c852f198/src/commitment/pedersen/mod.rs#L39).
It has [Parameters](https://github.com/arkworks-rs/crypto-primitives/blob/1e4f16cddba8d18f4eb219fa33ac25b2c852f198/src/commitment/pedersen/mod.rs#L40) thus they can change the commitment domain flexibly.

In microsoft, they do abstraction with [CommitmentEngineTrait](https://github.com/microsoft/Nova/blob/main/src/traits/commitment.rs#L76).
It has [CommitmentKey](https://github.com/microsoft/Nova/blob/main/src/traits/commitment.rs#L78) and [Commitment](https://github.com/microsoft/Nova/blob/main/src/traits/commitment.rs#L81) thus they have flexibility of range and domain.

In our implementation, we don't do abstraction and [Params](https://github.com/privacy-scaling-explorations/folding-schemes/blob/main/src/pedersen.rs#L17) corresponds to [Pedersen](https://github.com/privacy-scaling-explorations/folding-schemes/blob/main/src/pedersen.rs#L23) as one-to-one.
In that case, Pedersen should have params necessary for commitment in its fields.

Do you plan to support other commitment schemes and do abstraction?
I am going to change accordingly.

Thank you.

arnaucube: We discussed it with @CPerezz , and we agree to keep with the current design (previous to this PR) for Pedersen, since is a bit more 'future-proof' in the sense that it already follows the style of other arkworks components, and if eventually we need to allow other commitment schemes we don't need to undo the commit from the current PR.
So, if it's fine for you, if you remove the commit https://github.com/privacy-scaling-explorations/folding-schemes/pull/28/commits/22343757b0c6a9f322750527e07b65c99b29c7b5 from the PR, we're happy to include the other changes from this PR :)

ashWhiteHat: Cool!
I modified my PR accordingly.
I would appreciate it if you could confirm.

arnaucube: Awesome, thanks a lot!

arnaucube (approved):
LGTM! Thanks for the contribution!

Tagging also @CPerezz to review it.

CPerezz (approved):
LGTM

PR #29, opened 2023-09-28
Implement NIFS.P & NIFS.V transcript

ashWhiteHat:
Hi there
I replaced mocked transcript randomness with PoseidonTranscript.

I would appreciate it if you could confirm.
Thank you.

### DIFF
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -1,1 +1,2 @@
+pub(crate) const NUM_CHALLENGE_BITS: usize = 128;
 pub const N_BITS_CHALLENGE: usize = 250;
--- a/src/folding/nova/nifs.rs
+++ b/src/folding/nova/nifs.rs
@@ -1,12 +1,15 @@
+use ark_crypto_primitives::sponge::poseidon::PoseidonConfig;
 use ark_crypto_primitives::sponge::Absorb;
 use ark_ec::{CurveGroup, Group};
+use ark_ff::{BigInteger, PrimeField};
 use ark_std::One;
 use std::marker::PhantomData;
 
 use crate::ccs::r1cs::R1CS;
+use crate::constants::NUM_CHALLENGE_BITS;
 use crate::folding::nova::{CommittedInstance, Witness};
 use crate::pedersen::{Params as PedersenParams, Pedersen, Proof as PedersenProof};
-use crate::transcript::Transcript;
+use crate::transcript::{poseidon::PoseidonTranscript, Transcript};
 use crate::utils::vec::*;
 
 /// Implements the Non-Interactive Folding Scheme described in section 4 of
@@ -88,8 +91,8 @@ where
     #[allow(clippy::type_complexity)]
     pub fn prove(
         pedersen_params: &PedersenParams<C>,
-        // r comes from the transcript, and is a n-bit (N_BITS_CHALLENGE) element
-        r: C::ScalarField,
+        poseidon_config: PoseidonConfig<C::ScalarField>,
+        pp_digest: C::ScalarField,
         r1cs: &R1CS<C::ScalarField>,
         w1: &Witness<C>,
         ci1: &CommittedInstance<C>,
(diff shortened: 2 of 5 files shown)


src/folding/nova/nifs.rs
```diff
@@ -115,12 +133,27 @@ where
 
     // NIFS.V
     pub fn verify(
-        // r comes from the transcript, and is a n-bit (N_BITS_CHALLENGE) element
-        r: C::ScalarField,
+        poseidon_config: PoseidonConfig<C::ScalarField>,
```
arnaucube, line 136: Here the idea was that `r` comes from the transcript which is generated outside (previous) to this method call.

src/folding/nova/nifs.rs
```diff
@@ -115,12 +133,27 @@ where
...
+        pp_digest: C::ScalarField,
         ci1: &CommittedInstance<C>,
         ci2: &CommittedInstance<C>,
         cmT: &C,
     ) -> CommittedInstance<C> {
+        // initialize a new poseidon transcript
+        let mut tr = PoseidonTranscript::<C>::new(&poseidon_config);
```
arnaucube, line 143: The transcript will be started previous to the call of this function.

src/constants.rs
```diff
@@ -1 +1,2 @@
+pub(crate) const NUM_CHALLENGE_BITS: usize = 128;
 pub const N_BITS_CHALLENGE: usize = 250;
```
arnaucube, line 2: These 2 constants seem to have the same name (with different order), which is the difference between them?

src/folding/nova/nifs.rs
```diff
@@ -214,19 +266,41 @@ mod tests {
...
+        );
 
+        // compute a challenge
+        let r = compute_challenge::<Projective>(
+            poseidon_config,
+            <Projective as Group>::ScalarField::ZERO,
+            &ci1,
+            &ci2,
+            &cmT,
+        );
```
arnaucube, line 303: This was done with a random generated `r` instead of a full trancript building & appending, since this is a test and we don't need to recreate the full transcript, which adds complexity to the logic of the test and extra lines to read.

src/folding/nova/nifs.rs
```diff
@@ -308,12 +382,14 @@ mod tests {
...
-            let r = Fr::rand(&mut rng); // folding challenge would come from the transcript
+            let poseidon_config = poseidon_test_config::<Fr>();
+            let pp_digest = <Projective as Group>::ScalarField::ZERO;
 
             // NIFS.P
             let (folded_w, _, _, cmT) = NIFS::<Projective>::prove(
                 &pedersen_params,
-                r,
+                poseidon_config.clone(),
+                pp_digest,
```
arnaucube, line 392: Same as commented at line 143.

arnaucube (commented):
I'm not sure I understand the intention of this PR. From what I see, it modifies the NIFS.{prove/verify} methods to take `poseidon_config` and build a transcript inside instead of taking the `r` challenge as input.
But the idea was that `r` is given as parameter since is generated by the transcript that lives previously to these methods call, as was stated in the comment of line [`nifs.rs#L91`](https://github.com/privacy-scaling-explorations/folding-schemes/blob/8256c27609a2effbf9a255f0709eb0b1c3055b36/src/folding/nova/nifs.rs#L91):
`// r comes from the transcript, and is a n-bit (N_BITS_CHALLENGE) element`

This PR seems to be recreating microsoft/Nova's [nifs.rs prove & verify logic](https://github.com/microsoft/Nova/blob/fb83e30e16e56b3b21c519b15b83c4ce1f077a13/src/nifs.rs#L36), passing the `ro_consts`/`poseidon_config` & `pp_digest`, but the current repo `folding-schemes` is not intended to rewrite microsoft/Nova's in arkworks but to do a independent (different) implementation of Nova scheme (and other folding schemes too).

As in the other opened PRs, I'm sorry for the delayed replies since I'm OOO with low internet connection.

ashWhiteHat:
> the current repo folding-schemes is not intended to rewrite microsoft/Nova's in arkworks but to do a independent (different) implementation of Nova scheme (and other folding schemes too).
> The transcript will be started previous to the call of this function.

Thank you for polite explanation!
That makes sense.
Sorry this PR is mistake.

PR #30, opened 2023-10-03
add error handling to several pending methods

arnaucube:
Add error handling to several pending methods.
So with this PR for example, the [test from this other PR/branch](https://github.com/yugonsan/folding-schemes/blob/877ee39a68289ec7c040861bd726e00304c1d1fc/src/frontend/circom/mod.rs#L208) it will not panic failing internally in the call  of `check_relation` before the `expect`, and the test can be modified to expect the specific error.


han0110 (approved):
LGTM!

CPerezz (approved):
LGTM!

PR #39, opened 2023-11-28
add README.md

arnaucube:
PR to iterate the README.md to work on it and have it ready when Nova+CycleFold is ready.


.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "rust-analyzer.cargo.extraEnv": {
+        "LIBCLANG_PATH": "/Library/Developer/CommandLineTools/usr/lib/"
+    }
+}
```
arnaucube, line 5: I don't know if it's possible, but maybe there is a way to locally configure vscode to prevent this from being added? In a way that we can keep the repo code-editor-agnostic

README.md
```diff
@@ -37,15 +37,63 @@ Available frontends to define the folded circuit:
...
 - [Circom](https://github.com/iden3/circom), iden3, 0Kims Association
 
-
 ## Usage
 
+Detailed usage and design documentation can be found [here](https://privacy-scaling-explorations.github.io/sonobe-docs/).
+
+With sonobe, developers can choose which folding scheme and decider to use. The following code snippet shows how to fold a circuit using the Nova folding scheme and an EVM-compatible decider:
+
+```rust
```
arnaucube, line 46: I'm unsure on having a partial example in this readme since we moved the code samples into [sonobe-docs](https://privacy-scaling-explorations.github.io/sonobe-docs/usage/overview.html).

README.md
```diff
@@ -37,15 +37,63 @@ Available frontends to define the folded circuit:
...
+}
+```
+
+See the full example for using sonobe to recursively check bitcoin blocks [here](https://github.com/dmpierre/folding-schemes-light-btc/tree/main).
+
+### Folding Schemes and IVC
+
+A folding scheme reduces the task of checking multiple instances in some relation to the task of checking a single one. The canonical definition of a folding scheme entailed computing a folded instance as a random linear combination of the original instances. 
+
+A folding scheme can also be used as a primitive to achieve incrementally verifiable computation (IVC). One such example is [Nova](https://eprint.iacr.org/2021/370.pdf), which showed an IVC scheme with interesting performance properties. Note that an IVC proof is neither succint nor zero-knowledge. Hence, a final ("decider") zkSNARK will be required to prove the correctness of an IVC result.
```
arnaucube, line 88: maybe this is a 'duplicate' of the other "Folding Schemes introduction" section below? not that needs to be removed but maybe we could merge both of them into a single one?

PR #51, opened 2024-01-05
feat: `EqEvalGadget` for computing `eq(x, y)`

dmpierre:
A gadget for computing $\tilde{eq}(x, y)$. We will use this gadget within the hypernova nimfs verifier.

### DIFF
--- a/src/folding/circuits/nimfs_verifier.rs
+++ b/src/folding/circuits/nimfs_verifier.rs
@@ -1,2 +1,78 @@
 // hypernova nimfs verifier circuit
 // see section 5 in https://eprint.iacr.org/2023/573.pdf
+use std::marker::PhantomData;
+
+use ark_ff::PrimeField;
+use ark_r1cs_std::fields::{fp::FpVar, FieldVar};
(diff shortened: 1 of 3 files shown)


src/folding/circuits/nimfs_verifier.rs
```diff
@@ -1,2 +1,78 @@
 // hypernova nimfs verifier circuit
 // see section 5 in https://eprint.iacr.org/2023/573.pdf
```
arnaucube, line 2: I agree with having the `EqEvalGadget` in `src/folding/circuits`, since it will be reused also by some versions of ProtoGalaxy, but this directory should be agnostic to hypernova (thus not containing references to HyperNova's NIMFS).
How would you see to change the file name into something like `src/folding/circuits/utils.rs` so it can contain other utilities in the future, but the future `nimfs` specific gadgets would be at `src/folding/hypernova/circuits.rs`?

dmpierre: Yes definitely!

arnaucube (approved):
LGTM

PR #53, opened 2024-01-09
A gadget to compute the first term of the sum in section 5, step 5

dmpierre:
A gadget to compute the sum of all $\gamma^{j} \cdot eq(r_{x_j}, r_x^{\prime}) \cdot \sigma_j$.

Refactored `compute_c_from_sigmas_and_thetas` to properly test gadget.


src/folding/circuits/hypernova/utils.rs
```diff
@@ -0,0 +1,129 @@
...
+        self.value.pow_le(&pow)
+    }
+}
+
+/// Gadget to compute the sum of all $\gamma^{j} \cdot eq(r_{x_j}, r_x^{\prime}) \cdot \sigma_j$.
+pub struct SumMulsGammaPowsEqSigmaGadget<F: PrimeField> {
+    _f: PhantomData<F>,
+}
+
+impl<F: PrimeField> SumMulsGammaPowsEqSigmaGadget<F> {
```
dmpierre, line 40: Gadget to compute $\Sigma_{j}^{j + n} \gamma^{j} \cdot eq_eval \cdot \sigma_{j}$, where $n$ is the length of the `sigmas` vector and $j$ corresponds to `ccs.t` in `compute_c_from_sigmas_and_thetas`

src/folding/hypernova/utils.rs
```diff
@@ -86,6 +86,22 @@ pub fn compute_sigmas_and_thetas<C: CurveGroup>(
     SigmasThetas(sigmas, thetas)
 }
 
+/// Computes the sum $\Sigma_{j = 0}^{n} \gamma^{\text{pow} + j} \cdot eq_eval \cdot \sigma_{j}$
+/// `pow` corresponds to `i * ccs.t` in `compute_c_from_sigmas_and_thetas`
+pub fn sum_muls_gamma_pows_eq_sigma<F: PrimeField>(
```
dmpierre, line 91: Reference function to which we compare the gadget

src/folding/circuits/hypernova/utils.rs
```diff
@@ -0,0 +1,129 @@
...
+            let eq_var = FpVar::<Fr>::new_constant(cs.clone(), e_lcccs[i]).unwrap();
+            let pow = FpVar::<Fr>::new_constant(cs.clone(), Fr::from((i * ccs.t) as u64)).unwrap();
+            let computed = SumMulsGammaPowsEqSigmaGadget::sum_muls_gamma_pows_eq_sigma(
+                sigmas_var,
+                eq_var,
+                gamma_var.clone(),
+                pow,
+            )
+            .unwrap();
+            assert_eq!(expected, computed.value().unwrap());
```
dmpierre, line 126: Test is a bit messy since there are a few pieces that need to be initialized.

src/folding/circuits/hypernova/utils.rs
```diff
@@ -0,0 +1,129 @@
...
+    alloc::{AllocVar, AllocationMode},
+    fields::{fp::FpVar, FieldVar},
+    ToBitsGadget,
+};
+use ark_relations::r1cs::{Namespace, SynthesisError};
+use std::{borrow::Borrow, marker::PhantomData};
+
+/// `GammaVar` is a wrapper around an `FpVar` for $\gamma$.
+#[derive(Clone)]
+pub struct GammaVar<F: PrimeField> {
```
arnaucube, line 13: How would you see to move all the circuit related types and methods into `hypernova/circuits.rs`?

src/folding/circuits/hypernova/utils.rs
```diff
@@ -0,0 +1,129 @@
...
+        j: FpVar<F>,
+    ) -> Result<FpVar<F>, SynthesisError> {
+        let mut result = FpVar::<F>::zero();
+        let mut gamma_pow = gamma.pow(j).map_err(|_| SynthesisError::Unsatisfiable)?;
+        for sigma in sigmas {
+            result += gamma_pow.clone() * eq_eval.clone() * sigma;
+            gamma_pow *= gamma.value.clone();
+        }
+        Ok(result)
+    }
```
arnaucube, line 65: I think that the `sum_muls_gamma_pows_eq_sigma` method can be done without the `GammaVar` type wraper, being easier to follow, and getting rid of the GammaVar related code (lines 11 to 33). How would you see it?

It would be something like this, plus removing the lines from 11 to 33:
```suggestion
    pub fn sum_muls_gamma_pows_eq_sigma(
        sigmas: VecFpVar<F>,
        eq_eval: FpVar<F>,
        gamma: FpVar<F>,
        j: FpVar<F>,
    ) -> Result<FpVar<F>, SynthesisError> {
        let mut result = FpVar::<F>::zero();
        let mut gamma_pow = gamma.pow_le(&j.to_bits_le()?)?;
        for sigma in sigmas {
            result += gamma_pow.clone() * eq_eval.clone() * sigma;
            gamma_pow *= gamma.clone();
        }
        Ok(result)
    }
```


(For the pow logic, is like when we do scalar multiplication inside a circuit, we do `point.scalar_mul_le(bits.iter()?)`, we could add a type wrapper for that and add custom methods, but since it can already be done with a single line I think that it is easier to follow rather than adding extra calls to custom methods to extra type wrappers.)

src/folding/circuits/hypernova/utils.rs
```diff
@@ -0,0 +1,129 @@
...
+        // Initialize cs and gamma
+        let cs = ConstraintSystem::<Fr>::new_ref();
+        let gamma_var = GammaVar::<Fr>::new_constant(cs.clone(), gamma).unwrap();
+
+        for (i, sigmas) in sigmas_thetas.0.iter().enumerate() {
+            let expected =
+                sum_muls_gamma_pows_eq_sigma(gamma, e_lcccs[i], sigmas, (i * ccs.t) as u64);
+            let sigmas_var = VecFpVar::<Fr>::new_constant(cs.clone(), sigmas).unwrap();
+            let eq_var = FpVar::<Fr>::new_constant(cs.clone(), e_lcccs[i]).unwrap();
+            let pow = FpVar::<Fr>::new_constant(cs.clone(), Fr::from((i * ccs.t) as u64)).unwrap();
```
arnaucube, line 118: I understand that in the final circuit those are not going to be constants, right? Maybe the test could already use `new_witness`?

src/folding/circuits/utils.rs
```diff
@@ -1,7 +1,46 @@
...
+                .iter()
+                .map(|element| {
+                    FpVar::new_variable(cs.clone(), || Ok(element), mode)
+                        .map_err(|_| SynthesisError::Unsatisfiable)
+                })
+                .collect::<Result<Vec<_>, _>>()?;
+            Ok(Self { vec })
+        })
+    }
+}
```
arnaucube, line 43: I'm not sure to understand the reason to add the `VecFpVar` type wrapper, could we not just directly use `Vec<FpVar<F>>` when needed?
In this way we can avoid the lines 9 to 43 (getting rid of the `VecFpVar` wrapper and methods)

arnaucube (commented):
Nice work!! Left some comments

src/folding/circuits/hypernova/utils.rs
```diff
@@ -0,0 +1,129 @@
...
+    }
+}
+
+impl<F: PrimeField> GammaVar<F> {
+    /// An helper to compute $\gamma^{pow}$
+    pub fn pow(&self, pow: FpVar<F>) -> Result<FpVar<F>, SynthesisError> {
+        let pow = pow.to_bits_le()?;
+        self.value.pow_le(&pow)
+    }
+}
```
arnaucube, line 33: As mentioned in https://github.com/privacy-scaling-explorations/folding-schemes/pull/53/files#r1446223413 , I'm not sure to understand the reason to add the `GammaVar` type wrapper, the proposal would be to get rid of `GammaVar` and its methods.

src/folding/circuits/hypernova/utils.rs
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        // Initialize cs and gamma
+        let cs = ConstraintSystem::<Fr>::new_ref();
+        let gamma_var = GammaVar::<Fr>::new_constant(cs.clone(), gamma).unwrap();
+
+        for (i, sigmas) in sigmas_thetas.0.iter().enumerate() {
+            let expected =
+                sum_muls_gamma_pows_eq_sigma(gamma, e_lcccs[i], sigmas, (i * ccs.t) as u64);
+            let sigmas_var = VecFpVar::<Fr>::new_constant(cs.clone(), sigmas).unwrap();
```
arnaucube, line 116: With the change from https://github.com/privacy-scaling-explorations/folding-schemes/pull/53/files#r1446250661 , this line would become something like
```suggestion
            let sigmas_var: Vec<FpVar<Fr>> = Vec::new_constant(cs.clone(), sigmas.clone()).unwrap();
```

src/folding/circuits/utils.rs
dmpierre: Later on, we will want to compute the product and also the sum of elements of a vector of `FpVar`- namely for the second part of the final sum. This abstraction helps us then do to things like:

```rust
for c_i in c {
     // ...
     let pi_thetaj = thetas[i].product()?;
     //. ...
}
// [...]
let sum = VecFpVar::new(result_vec).sum()?;
```

Also vectors of `FpVar` should be relatively common in other schemes? I think this abstraction might be helpful down the road.

That was my rationale for doing this. I might be wrong though, so open to other ways of doing things. 

What do you think?

src/folding/circuits/hypernova/utils.rs
dmpierre: [Done](https://github.com/privacy-scaling-explorations/folding-schemes/pull/53/commits/e3fcb8790d790c25267fcdfb40e820bf76a6aa3e)!

src/folding/circuits/hypernova/utils.rs
dmpierre: [Yup](https://github.com/privacy-scaling-explorations/folding-schemes/pull/53/commits/e3fcb8790d790c25267fcdfb40e820bf76a6aa3e)!

src/folding/circuits/hypernova/utils.rs
dmpierre: [Yes!](https://github.com/privacy-scaling-explorations/folding-schemes/pull/53/commits/005efd1e1448014ef6e8f74a92a41ced10eff6ba)

src/folding/circuits/utils.rs
arnaucube: I'm not sure, the same logic could be said for doing type wrappers for `Vec<Fr>` (in native rust, outside circuits), but we don't add a wrapper there.

The two examples from the previous comment (sum & prod) can be done without type wrappers in one line just with things like `.iter()`:

```rust
    #[test]
    fn test_iter() {
        let v = vec![
            Fr::from(1u32),
            Fr::from(2u32),
            Fr::from(3u32),
            Fr::from(4u32),
        ];

        // in rust natively over Vec<Fr>
        let sum: Fr = v.iter().sum();
        let prod: Fr = v.iter().product();

        let vVar = Vec::<FpVar<Fr>>::new_witness(cs.clone(), || Ok(v.clone())).unwrap();

        // inside circuit over Vec<FpVar<Fr>>
        let sumVar: FpVar<Fr> = vVar.iter().sum();
        let prodVar: FpVar<Fr> = vVar.iter().fold(FpVar::one(), |acc, e| acc * e);

        assert_eq!(vVar.value().unwrap(), v);
        assert_eq!(sumVar.value().unwrap(), sum);
        assert_eq!(prodVar.value().unwrap(), prod);
    }
```

src/folding/circuits/hypernova/utils.rs
arnaucube: I don't know why, but I can see the move happening in the commit that you link, but if I go to the last PR changes it appears still under `src/folding/circuits/hypernova/utils.rs`
and also if I go to the latest commit of the branch [dev/hypernova-nimfs-verifier-circuit/mul-gammapow-eq-sigma-gadget](https://github.com/privacy-scaling-explorations/folding-schemes/tree/dev/hypernova-nimfs-verifier-circuit/mul-gammapow-eq-sigma-gadget) :eyes:

arnaucube: Oh I see what's happening, the `SumMulsGammaPowsEqSigmaGadget` gadget is duplicated at
https://github.com/privacy-scaling-explorations/folding-schemes/blob/005efd1e1448014ef6e8f74a92a41ced10eff6ba/src/folding/hypernova/circuit.rs#L18
and https://github.com/privacy-scaling-explorations/folding-schemes/blob/005efd1e1448014ef6e8f74a92a41ced10eff6ba/src/folding/circuits/hypernova/utils.rs#L40

dmpierre: Ah thanks, I think I rebased incorrectly.. [Done](https://github.com/privacy-scaling-explorations/folding-schemes/pull/53/commits/b939374c62ad2dd4c09f329b59ffcb6820e201ab)

src/folding/circuits/utils.rs
dmpierre: Thanks, I was not aware that you could call `Vec::<FpVar<Fr>>::new_witness`! Really cool. Removed the `VecFpVar` struct [here](https://github.com/privacy-scaling-explorations/folding-schemes/pull/53/commits/5c48096c995dbd07ffe44a9b6a36f16c4e7e5c28)

src/folding/circuits/hypernova/utils.rs
dmpierre: Yup [done](https://github.com/privacy-scaling-explorations/folding-schemes/pull/53/commits/5c48096c995dbd07ffe44a9b6a36f16c4e7e5c28)

src/folding/hypernova/circuit.rs
```diff
@@ -1,2 +1,105 @@
...
+}
+
+impl<F: PrimeField> SumMulsGammaPowsEqSigmaGadget<F> {
+    /// Computes the sum $\Sigma_{j}^{j + n} \gamma^{j} \cdot eq_eval \cdot \sigma_{j}$, where $n$ is the length of the `sigmas` vector
+    /// It corresponds to the first term of the sum that $\mathcal{V}$ has to compute at section 5, step 5 of "A multi-folding scheme for CCS".
+    ///
+    /// # Arguments
+    /// - `sigmas`: vector of $\sigma_j$ values
+    /// - `eq_eval`: the value of $\tilde{eq}(x_j, x^{\prime})$
+    /// - `gamma`: a `GammaVar`, which supports a `pow` method, representing $\gamma$
```
arnaucube, line 24: Maybe now the doc for `gamma` can be updated to remove the `GammaVar` reference

dmpierre: Ah good catch! [Done]([06806c8](https://github.com/privacy-scaling-explorations/folding-schemes/pull/53/commits/06806c8d4fb686ee31e4b52bb3eace90a6902647))

arnaucube (approved):
Nice work! ^^

PR #56, opened 2024-01-10
A gadget to compute the second term of the sum in section 5, step 5 

dmpierre:
Gadget which computes $\Sigma_{i=1}^{q} c_i \cdot \Pi_{j \in S_i} \theta_j$. 

Will then multiply this term by $\gamma^{t+1}$ and $e_2$ and sum it with the first part of the sum to obtain the final gadget!

### DIFF
--- a/src/folding/hypernova/circuit.rs
+++ b/src/folding/hypernova/circuit.rs
@@ -74,7 +74,7 @@ impl<F: PrimeField> SumCiMulProdThetajGadget<F> {
 /// An helper function to run before computing $\Pi_{j \in S_i} \theta_j$ in a circuit.
 pub fn get_prepared_thetas<C: CurveGroup>(
     S_i: &Vec<usize>,
-    thetas: &Vec<C::ScalarField>,
+    thetas: &[C::ScalarField],
 ) -> Vec<C::ScalarField> {
     let mut prepared: Vec<C::ScalarField> = Vec::new();
     for j in S_i {
--- a/src/folding/hypernova/circuit.rs
+++ b/src/folding/hypernova/circuit.rs
@@ -21,7 +21,7 @@ impl<F: PrimeField> SumMulsGammaPowsEqSigmaGadget<F> {
     /// # Arguments
     /// - `sigmas`: vector of $\sigma_j$ values
     /// - `eq_eval`: the value of $\tilde{eq}(x_j, x^{\prime})$
-    /// - `gamma`: a `GammaVar`, which supports a `pow` method, representing $\gamma$
+    /// - `gamma`: value $\gamma$
     /// - `j`: the power at which we start to compute $\gamma^{j}$. This is needed in the contexxt of multifolding.
     ///
     /// # Notes
--- a/src/folding/hypernova/circuit.rs
+++ b/src/folding/hypernova/circuit.rs
@@ -73,11 +73,11 @@ impl<F: PrimeField> SumCiMulProdThetajGadget<F> {
 /// Returns a vector of thetas for a corresponding $S_i$
 /// An helper function to run before computing $\Pi_{j \in S_i} \theta_j$ in a circuit.
 pub fn get_prepared_thetas<C: CurveGroup>(
-    S: &Vec<usize>,
+    S_i: &Vec<usize>,
     thetas: &Vec<C::ScalarField>,
 ) -> Vec<C::ScalarField> {
     let mut prepared: Vec<C::ScalarField> = Vec::new();
-    for j in S {
+    for j in S_i {
         prepared.push(thetas[*j]);
     }
     prepared
--- a/src/folding/hypernova/circuit.rs
+++ b/src/folding/hypernova/circuit.rs
@@ -42,15 +42,44 @@ impl<F: PrimeField> SumMulsGammaPowsEqSigmaGadget<F> {
     }
 }
 
+/// Gadget to compute $\Sigma_{i=1}^{q} c_i * \Pi_{j \in S_i} theta_j$.
+pub struct SumCiMulProdThetajGadget<F: PrimeField> {
+    _f: PhantomData<F>,
+}
+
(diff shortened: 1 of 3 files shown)


src/folding/hypernova/circuit.rs
```diff
@@ -21,7 +22,7 @@ impl<F: PrimeField> SumMulsGammaPowsEqSigmaGadget<F> {
     /// # Arguments
     /// - `sigmas`: vector of $\sigma_j$ values
     /// - `eq_eval`: the value of $\tilde{eq}(x_j, x^{\prime})$
-    /// - `gamma`: a `GammaVar`, which supports a `pow` method, representing $\gamma$
+    /// - `gamma`: value $\gamma$
```
arnaucube, line 25: I'm surprised to see this change, I thought that this was already done in the previous PR and already included in the pointed branch `feature/hypernova-nimfs-verifier-circuit` (https://github.com/privacy-scaling-explorations/folding-schemes/blob/11066b15f9ef4c97dba2b05608b0d093405ae0f7/src/folding/hypernova/circuit.rs#L24).
Is the branch of the current PR not departing from latest version of `feature/hypernova-nimfs-verifier-circuit`?

arnaucube: to be clear, this is not a request of change, just that I'm curious with the current diff hehe

src/folding/hypernova/circuit.rs
```diff
@@ -42,15 +43,60 @@ impl<F: PrimeField> SumMulsGammaPowsEqSigmaGadget<F> {
...
+            let prod = &thetas[i].iter().fold(FpVar::one(), |acc, e| acc * e);
+            result += c_i * prod;
+        }
+        result
+    }
+}
+
+/// Returns a vector of thetas for a corresponding $S_i$
+/// An helper function to run before computing $\Pi_{j \in S_i} \theta_j$ in a circuit.
+pub fn get_prepared_thetas<C: CurveGroup>(
```
arnaucube, line 75: Currently this method is only used in the tests, I understand that in a future iteration it will be used by some non-test method, right?
(if not, I would suggest to move it inside the tests)

Additionally, I'm wondering that since the method just does a loop & push, maybe it could be easier to read the code without jumping between methods and directly doing the loop computation inline (something like a oneline iterator doing the push without requiring a whole new method). How do you see it?

src/folding/hypernova/circuit.rs
dmpierre: I rebased feature/hypernova-nimfs-verifier-circuit` twice on this branch before and after the merge, this might explain why the change for this updated comment is here! Sorry, this is kinda confusing.

arnaucube: oh cool, make sense. Was asking to learn, thanks ^^

src/folding/hypernova/circuit.rs
```diff
@@ -102,4 +148,44 @@ mod tests {
...
+
+        // Initialize cs
+        let cs = ConstraintSystem::<Fr>::new_ref();
+        let vec_thetas = sigmas_thetas.1;
+        for (_, thetas) in vec_thetas.iter().enumerate() {
+            // sum c_i * prod theta_j
+            let expected = sum_ci_mul_prod_thetaj(&ccs, thetas); // from `compute_c_from_sigmas_and_thetas`
+            let mut prepared_thetas = Vec::new();
+            for i in 0..ccs.q {
+                let prepared = get_prepared_thetas::<Projective>(&ccs.S[i], thetas);
```
arnaucube, line 180: in relation to the comment https://github.com/privacy-scaling-explorations/folding-schemes/pull/56/files#r1447607881
looks like this line would be achieved without the extra method `get_prepared_thetas` by doing something like
```suggestion
                let prepared: Vec<Fr> = ccs.S[i].iter().map(|j| thetas[*j]).collect();
```

src/folding/hypernova/circuit.rs
```diff
@@ -42,15 +43,60 @@ impl<F: PrimeField> SumMulsGammaPowsEqSigmaGadget<F> {
     }
 }
 
+/// Gadget to compute $\Sigma_{i=1}^{q} c_i * \Pi_{j \in S_i} theta_j$.
+pub struct SumCiMulProdThetajGadget<F: PrimeField> {
+    _f: PhantomData<F>,
+}
+
+impl<F: PrimeField> SumCiMulProdThetajGadget<F> {
+    /// Computes $\Sigma_{i=1}^{q} c_i * \Pi_{j \in S_i} theta_j$
```
arnaucube, line 52: just a small thing that I just noticed (applies to the other part of code done in the previous PRs)
usually
`$\Sigma,~ \Pi$`: ($\Sigma, \Pi$) are used for values,
whereas for the sum and product can use `$\sum, \prod$`: ($\sum,~ \prod$)

But it's just a very small 'stylish' thing that does not affect the code, so if you prefer to use `\Sigma, \Pi` it's fine too.

arnaucube (commented):
Great! Getting closer to the composition of the methods to get the verifier ^^

Just some small comments

src/folding/hypernova/circuit.rs
dmpierre: Yes def, thanks for the heads up! Changed at [bdfd7bd](https://github.com/privacy-scaling-explorations/folding-schemes/pull/56/commits/bdfd7bd93ee38e20f06498581f517ad243e4d6f2)

src/folding/hypernova/circuit.rs
dmpierre: Yup I agree, updated with [ac92a8b](https://github.com/privacy-scaling-explorations/folding-schemes/pull/56/commits/ac92a8b38aae2a58964000c77805286b635f8a38)

arnaucube (approved):
:cat: :rocket:

PR #63, opened 2024-01-26
Adapt Nova usage of commitment to the generic trait

arnaucube:


src/folding/nova/decider.rs
```diff
@@ -216,25 +220,33 @@ where
...
+impl<C1, GC1, C2, GC2, CP1, CP2> DeciderCircuit<C1, GC1, C2, GC2, CP1, CP2>
 where
     C1: CurveGroup,
     C2: CurveGroup,
     GC1: CurveVar<C1, CF2<C1>>,
     GC2: CurveVar<C2, CF2<C2>>,
+    CP1: CommitmentProver<C1>,
+    CP2: CommitmentProver<C2>,
+    // enforce that the CP2 is Pedersen commitment
+    PedersenParams<C2>: From<<CP2 as CommitmentProver<C2>>::Params>,
```
han0110, line 232: Perhaps just fix the `Params` directly? Or if there is a chance the param is in different struct then it's fine to keep as is.
```suggestion
    CP2: CommitmentProver<C2, Params = PedersenParams<C2>>,
    // enforce that the CP2 is Pedersen commitment
```

han0110 (approved):
LGTM! Just a tiny question.

src/folding/nova/decider.rs
arnaucube: Oh right! [Updated](https://github.com/privacy-scaling-explorations/folding-schemes/compare/57180c01340c50ca73c96c3d95a1115f87443fb1..9590ad31bc05965959216a673ac139c986354387), thanks!

rust-toolchain
```diff
@@ -1 +1 @@
-1.73.0
+1.75.0
```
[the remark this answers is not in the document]
arnaucube: I was keeping with the latest release, but if it makes sense to stick to a specific one I'm also happy.

PR #69, opened 2024-02-07
Adding groth16, kzg10 and final decider verifiers in a dedicated workspace

dmpierre:
This PR re-organizes the repo into two different workspaces: `folding-schemes` and `folding-schemes-solidity`. 

While the `folding-schemes` workspace holds initial code, the `folding-schemes-solidity` one contains all relevant evm contracts for each of the verifiers. Three different contracts are included, along with their respective templates. We provide adequate tests for each of those.

- Templates are generated using the [askama]() library and can be found [here](https://github.com/privacy-scaling-explorations/folding-schemes/tree/feature/solidity-verifiers/folding-schemes-solidity/templates)
- Verifiers templates configuration are [here](https://github.com/privacy-scaling-explorations/folding-schemes/blob/feature/solidity-verifiers/folding-schemes-solidity/src/verifiers/templates.rs)
- We test that each of the verifiers templates, compiles and execute correctly [here](https://github.com/privacy-scaling-explorations/folding-schemes/blob/feature/solidity-verifiers/folding-schemes-solidity/src/verifiers/mod.rs). We do not yet include a test for the final decider, since additional details regarding how it will work are on hold.


folding-schemes/src/folding/nova/ivc.rs
```diff
@@ -0,0 +1 @@
+
```
arnaucube, line 1: this file does not exist in `main` branch https://github.com/privacy-scaling-explorations/folding-schemes/tree/main/src/folding/nova

folding-schemes/src/folding/nova/decider_eth_circuit.rs
```diff
@@ -0,0 +1,686 @@
+/// This file implements the onchain (Ethereum's EVM) decider circuit. For non-ethereum use cases,
```
arnaucube, line 1: this file appears as 'new' when it should be as 'file renamed without changes' (since it exist in the `main` branch: https://github.com/privacy-scaling-explorations/folding-schemes/blob/main/src/folding/nova/decider_eth_circuit.rs )
Could it be that the PR branch is not getting all the changes from the `main` branch?

arnaucube: Also for `pedersen.rs`. I've done in this branch an update getting last `main` version into `folding-schemes` subdir, if that is useful: https://github.com/privacy-scaling-explorations/folding-schemes/compare/main...feature/solidity-verifiers-from-last-main

dmpierre: Yes, I will close this PR and move to your branch. 

I've done a rebase early on, but on a branch different from main, to get your kzg implementation when testing. I think that's why we might end up with this! Sorry for the trouble :)

PR #70, opened 2024-02-07
Add solidity groth16, kzg10 and final decider verifiers in a dedicated workspace

dmpierre:
This PR re-organizes the repo into two different workspaces: folding-schemes and folding-schemes-solidity.

While the folding-schemes workspace holds initial code, the folding-schemes-solidity one contains all relevant evm contracts for each of the verifiers. Three different contracts are included, along with their respective templates. We provide adequate tests for each of those.

- Templates are generated using the askama library and can be found [here](https://github.com/privacy-scaling-explorations/folding-schemes/tree/feature/solidity-verifiers-from-last-main/folding-schemes-solidity/templates)
- Verifiers templates configuration are [here](https://github.com/privacy-scaling-explorations/folding-schemes/blob/feature/solidity-verifiers-from-last-main/folding-schemes-solidity/src/verifiers/templates.rs)
- We test that each of the verifiers templates, compiles and execute correctly [here](https://github.com/privacy-scaling-explorations/folding-schemes/blob/feature/solidity-verifiers-from-last-main/folding-schemes-solidity/src/verifiers/mod.rs). We do not yet include a test for the final decider, since additional details regarding how it will work are on hold.


folding-schemes-solidity/templates/kzg10_verifier.askama.sol
```diff
@@ -0,0 +1,263 @@
...
+     * @notice  Adds two points in G_1.
+     * @param   p1  G_1 point 1
+     * @param   p2  G_1 point 2
+     * @return  r  G_1 point p1 + p2
+     */
+    function add(uint256[2] memory p1, uint256[2] memory p2) internal view returns (uint256[2] memory r) {
+        bool success;
+        uint256[4] memory input = [p1[0], p1[1], p2[0], p2[1]];
+        assembly {
+            success := staticcall(sub(gas(), 2000), 6, input, 0xc0, r, 0x60)
```
han0110, line 55:
```suggestion
            success := staticcall(sub(gas(), 2000), 6, input, 0x80, r, 0x40)
```

folding-schemes-solidity/templates/kzg10_verifier.askama.sol
```diff
@@ -0,0 +1,263 @@
...
+
+        assembly {
+            success := staticcall(sub(gas(), 2000), 8, input, 0x180, out, 0x20)
+            switch success
+            case 0 { invalid() }
+        }
+
+        require(success, "bn254: pairing failed");
+
+        return out[0] != 0;
```
han0110, line 103: I feel `return out[0] == 1;` will be more straightforward, but it's fine as is.

han0110 (commented):
Overall LGTM! Left some nitpicks and I didn't review the `batchCheck` in `KZG10Verifier` because I didn't see how to proof is made and not sure how verifier should check.

folding-schemes-solidity/src/utils/encoding.rs
```diff
@@ -0,0 +1,45 @@
...
+#[derive(Debug, Default)]
+pub struct G1Repr(pub [FqWrapper; 2]);
+
+impl Display for G1Repr {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{:#?}", self.0)
+    }
+}
+
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
dmpierre: Ah yes, this is due to the fact that clippy would otherwise complain. Since those are only used within tests, clippy marks it as dead code. Do you think there could be some better way to achieve this?

folding-schemes-solidity/src/utils/encoding.rs
```diff
@@ -0,0 +1,45 @@
...
+#[derive(Debug, Default)]
+pub struct G2Repr(pub [[FqWrapper; 2]; 2]);
+
+impl Display for G2Repr {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{:#?}", self.0)
+    }
+}
+
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
dmpierre: see above comment!

folding-schemes-solidity/src/utils/encoding.rs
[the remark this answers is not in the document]
han0110: Ah I see, because I commented it but didn't found clippy warning, but the fact is the `Groth16Verifier::from` uses them but still marked with `#[allow(dead_code)]`.

I think if we expose the structure in `verifier::template` then it wouldn't be dead code anymore, for example in `lib.rs`:

```rust
pub use verifiers::templates::*;
```

Shouldn't these be exposed for further usage?

folding-schemes-solidity/templates/kzg10_verifier.askama.sol
dmpierre: good catch, thanks!

folding-schemes-solidity/templates/kzg10_verifier.askama.sol
```diff
@@ -0,0 +1,263 @@
...
+     * @return  r  G_1 point p multiplied by scalar s
+     */
+    function mulScalar(uint256[2] memory p, uint256 s) internal view returns (uint256[2] memory r) {
+        uint256[3] memory input;
+        input[0] = p[0];
+        input[1] = p[1];
+        input[2] = s;
+        bool success;
+        assembly {
+            success := staticcall(sub(gas(), 2000), 7, input, 0x80, r, 0x60)
```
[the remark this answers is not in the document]
dmpierre: this is better indeed, thank you!

folding-schemes-solidity/templates/kzg10_verifier.askama.sol
dmpierre: good suggestion! [478d9f2](https://github.com/privacy-scaling-explorations/folding-schemes/pull/70/commits/478d9f280ee1b8b5c201fed8730f03acf6ecbb6a)!

folding-schemes-solidity/src/utils/encoding.rs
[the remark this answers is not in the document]
dmpierre: Yes I think we can anticipate that those will be used, notably by the upcoming cli. This also results in cleaner code, so changed all that with [cb3561d](https://github.com/privacy-scaling-explorations/folding-schemes/pull/70/commits/cb3561dcebac4f7cb9b0bd6b0af1dd567c136fe4), exposing utilities for interacting with the evm and templates. Thanks for suggesting this!

han0110 (approved):
LGTM!

folding-schemes-solidity/src/utils/encoding.rs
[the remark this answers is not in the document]
han0110: I see, make sense

folding-schemes-solidity/templates/kzg10_verifier.askama.sol
```diff
@@ -0,0 +1,263 @@
+{{ sdpx }}
+
+{{ pragma_version }}
+
```
arnaucube, line 4: Should this contain some reference to the original code and its license?

folding-schemes-solidity/templates/groth16_verifier.askama.sol
```diff
@@ -0,0 +1,169 @@
+{{ sdpx }}
+/*
+    Copyright 2021 0KIMS association.
+
+    This file is generated with [snarkJS](https://github.com/iden3/snarkjs).
```
arnaucube, line 5: `This file is generated with snarkJS` --> is it? or is a fork of the template from snarkjs?
Maybe we it can keep the 0KIMS/iden3 license but updating the sentence? Maybe adding a link to the original template from snarkjs repo?
(I have not much idea on the licensing topic :S)

arnaucube (commented):
Great work!
Looks good, the only comments I left are regarding license / references, which I'm not sure on the topic.

folding-schemes-solidity/README.md
```diff
@@ -0,0 +1 @@
+TODO!!!
```
arnaucube, line 1: Maybe this for the moment can contain a sentence like
```
# folding-schemes-solidity
This crate implements the code to generate the Solidity code that verifies the folding-schemes decider proofs.
```

folding-schemes-solidity/templates/kzg10_verifier.askama.sol
dmpierre: Indeed, added a few comments on the code inspiration and the changes we did to it. See [8c843e5](https://github.com/privacy-scaling-explorations/folding-schemes/pull/70/commits/8c843e5a46ce4f7f739473f898b9d248390e3dc5)

The sdpx statement is templated because`solc` will not compile a solidity contract which contains conflicting licenses. In our case, we have the `kzg10` and `groth16` verifiers co-habitating for the final decider, one is licensed MIT and the other gpl-3. 

Templating allows us to just provide `None` for now and default to gpl-3 for the final decider contract, reminding ourselves that we should make a decision on that front!

folding-schemes-solidity/templates/groth16_verifier.askama.sol
dmpierre: I'm not too familiar with gpl-3 but I think that is ok? We should still figure out which and how the license should be included in general.
So I think that's a fair point and added this to the template with [8c843e5](https://github.com/privacy-scaling-explorations/folding-schemes/pull/70/commits/8c843e5a46ce4f7f739473f898b9d248390e3dc5)

arnaucube (approved):
Great!! :rocket:
Big step forward with this PR!

PR #73, opened 2024-02-21
Fix Nova multi-elements state

arnaucube:
@dmpierre found out that when using the multiple-elements state option for folding Nova it crashes. Turns out that in the AugmentedFCircuit the default value for the state when no input is provided was `vec![F::zero()]`, which defaults to length `1`. So when having more than 1 element in the state, before even starting to fold, just when creating the circuit, it already fails.

This PR fixes that, so now folding a state that has multiple elements works.

Also this PR adds an example for a circuit with a state of 5 elements.

### DIFF
--- a/folding-schemes/examples/multiple_inputs.rs
+++ b/folding-schemes/examples/multiple_inputs.rs
@@ -3,15 +3,9 @@
 #![allow(non_camel_case_types)]
 #![allow(clippy::upper_case_acronyms)]
 
-use ark_crypto_primitives::crh::{
-    sha256::{
-        constraints::{Sha256Gadget, UnitVar},
-        Sha256,
-    },
-    CRHScheme, CRHSchemeGadget,
-};
-use ark_ff::{BigInteger, PrimeField, ToConstraintField};
-use ark_r1cs_std::{fields::fp::FpVar, ToBytesGadget, ToConstraintFieldGadget};
+use ark_ff::PrimeField;
+use ark_r1cs_std::alloc::AllocVar;
+use ark_r1cs_std::fields::fp::FpVar;
 use ark_relations::r1cs::{ConstraintSystemRef, SynthesisError};
 use core::marker::PhantomData;
 use std::time::Instant;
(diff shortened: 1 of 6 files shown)


folding-schemes/examples/utils.rs
```diff
@@ -0,0 +1,48 @@
...
+
+use folding_schemes::commitment::pedersen::Pedersen;
+use folding_schemes::folding::nova::{get_r1cs, ProverParams, VerifierParams};
+use folding_schemes::frontend::FCircuit;
+use folding_schemes::transcript::poseidon::poseidon_test_config;
+
+// This method computes the Prover & Verifier parameters for the example. For a real world use case
+// those parameters should be generated carefuly (both the PoseidonConfig and the PedersenParams)
+#[allow(clippy::type_complexity)]
+pub(crate) fn nova_setup<FC: FCircuit<Fr>>(
```
dmpierre, line 17: I would use this name to emphasize that this should not be used in prod in any way.

```suggestion
pub(crate) fn test_nova_setup<FC: FCircuit<Fr>>(
```

folding-schemes/src/folding/nova/circuits.rs
```diff
@@ -310,10 +310,14 @@ where
             Ok(self.i.unwrap_or_else(CF1::<C1>::zero))
         })?;
         let z_0 = Vec::<FpVar<CF1<C1>>>::new_witness(cs.clone(), || {
-            Ok(self.z_0.unwrap_or(vec![CF1::<C1>::zero()]))
+            Ok(self
+                .z_0
+                .unwrap_or(vec![CF1::<C1>::zero(); self.F.state_len()]))
```
dmpierre, line 315: 🥳

dmpierre (approved):
A few comments, mostly nits and typos. My only question is how `state_len` is obtained,  abstracting this could be nice. Not sure how much work it would entail though, so I'm happy with it this way as well.
Thanks Arnau!

folding-schemes/examples/multiple_inputs.rs
```diff
@@ -0,0 +1,156 @@
...
+        let b = z_i[1].clone() + fourty.clone();
+        let c = z_i[2].clone() * four;
+        let d = z_i[3].clone() * fourty;
+        let e = z_i[4].clone() + onehundred;
+
+        Ok(vec![a, b, c, d, e])
+    }
+}
+
+/// cargo test --example multiple-inputs
```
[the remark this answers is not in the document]
dmpierre: Btw if you change the file name, this comment and the one after it will also need to be changed!

folding-schemes/examples/multiple_inputs.rs
```diff
@@ -0,0 +1,156 @@
+#![allow(non_snake_case)]
```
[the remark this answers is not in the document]
arnaucube: Right, in fact is not even _'add'_ because it does additions and multiplications, I've renamed it to `multi_inputs.rs` and the circuit is now `MultiInputsFCircuit`.

folding-schemes/examples/utils.rs
arnaucube: Adding the `test_` prefix makes it look as it is a test, but it is not a test but a function to be used in tests. It is in fact not a public function outside the crate, so users of the library should not have access to it.
But agree with you on emphasizing that this should not be used in production, added an extra 'warning' line in the comment of the function.

han0110 (approved):
LGTM!

folding-schemes/examples/utils.rs
arnaucube: Update: I had a second thought on it and changed my mind, even the function is not a test itself, I agree that is a priority to make it as more explicit as we can that is intended to be used only in tests. Updated here: https://github.com/privacy-scaling-explorations/folding-schemes/pull/73/commits/f64b13e05d87f2be7ef330230c7a6550a0558dad

PR #78, opened 2024-03-04
Add external inputs logic to F function/circuit. Add an example of usage with external inputs too.

arnaucube:


folding-schemes/examples/external_inputs.rs
```diff
@@ -0,0 +1,213 @@
...
+        i: usize,
+        z_i: Vec<FpVar<F>>,
+    ) -> Result<Vec<FpVar<F>>, SynthesisError> {
+        let crh_params =
+            CRHParametersVar::<F>::new_constant(cs.clone(), self.poseidon_config.clone())?;
+        let external_inputVar =
+            FpVar::<F>::new_witness(cs.clone(), || Ok(self.external_inputs[i])).unwrap();
+        let input: [FpVar<F>; 2] = [z_i[0].clone(), external_inputVar.clone()];
+        let h = CRHGadget::<F>::evaluate(&crh_params, &input)?;
+        Ok(vec![h, FpVar::<F>::zero()])
```
dmpierre, line 119: It seems like it might require some heavy lifting to avoid having to put a zero as the last element of the state. As we might have to keep things that way for now, I think it could be worth it to make an assert somewhere. Something like: 
```rust
 // in folding/nova/mod.rs, within `prove_step()`
if z_i1.last() != Some(&C1::ScalarField::zero()) {
    return Err(Error::MalformedZ_i); // `Err` value: z_i should have a zero as its last element
}
```

folding-schemes/src/lib.rs
dmpierre, line 1: Maybe add some other `Error` enum here regarding the expected format for the `z_i` state?

dmpierre (approved):
Lgtm! Thanks Arnau, happy to put my hands on this now :)

folding-schemes/examples/external_inputs.rs
```diff
@@ -0,0 +1,213 @@
+#![allow(non_snake_case)]
```
han0110, line 1: Just curious about the use case, this example could also be rewritten in the way the circuit handles the hash chain of signatures itself without an external input right?

han0110 (approved):
LGTM! Just a small question

folding-schemes/examples/external_inputs.rs
arnaucube: This is kind of a simple example using only hashes, but it showcases the approach that could be used for example to verify multiple signatures, one for each folding step. The FCircuit would check the signature and is folded for each of the signatures and public keys.
But: the signature is inputted from the external inputs, which are not part of the initial state.
So, following the [diagram from line 45](https://github.com/privacy-scaling-explorations/folding-schemes/pull/78/files#diff-0d0653ec716e47765490702600988ef72456e5a687824e0534afa0be0b8dd92dR45), the `w_i` "are not part" of the state, although internally they are used "as part" of the state to compute the new state.

The main difference in this PR is that now the FCircuit has access to know the `i` at each step internally (both at `step_native` and `generate_step_constraints`), which allows it to get specific values from some array, in the case of this example the circuit has this `external_inputs` array, which internally in the circuit are accessed at the `i` position.

I'm not sure if this is what you were referring to, does it make sense?

han0110: I see, thanks for explanation!

folding-schemes/examples/external_inputs.rs
dmpierre: Yup very clear, I see how this does not apply to other examples. Thanks for the detailed explanation! 

My rationale is that it is different from implementing the circuit itself, since this is solely about "formatting logic", independent of any in-circuit computation, solely enforcing that the output is in a digestible format for the subsequent folding circuit.

I'm not sure if I'm clear and I might be wrong, in any case this isn't even a blocking issue, rather a curious inquiry! Marking this as resolved :)

PR #81, opened 2024-03-27
 Link committed instances and r to the public input x in cyclefold circuit

### LINKED ISSUE #48: Link committed instances and r in cyclefold circuit

arnaucube:
resolves #48

### DIFF
--- a/folding-schemes/src/folding/circuits/nonnative.rs
+++ b/folding-schemes/src/folding/circuits/nonnative.rs
@@ -1,8 +1,7 @@
 use ark_ec::{AffineRepr, CurveGroup};
-use ark_r1cs_std::fields::nonnative::{params::OptimizationType, AllocatedNonNativeFieldVar};
 use ark_r1cs_std::{
     alloc::{AllocVar, AllocationMode},
-    fields::nonnative::NonNativeFieldVar,
+    fields::nonnative::{params::OptimizationType, AllocatedNonNativeFieldVar, NonNativeFieldVar},
 };
 use ark_relations::r1cs::{Namespace, SynthesisError};
 use ark_std::{One, Zero};
--- a/folding-schemes/src/folding/nova/circuits.rs
+++ b/folding-schemes/src/folding/nova/circuits.rs
@@ -251,6 +251,7 @@ pub struct AugmentedFCircuit<
     pub u_i: Option<CommittedInstance<C1>>,
     pub U_i: Option<CommittedInstance<C1>>,
     pub U_i1: Option<CommittedInstance<C1>>,
+    pub r_nonnat: Option<CF2<C1>>,
     pub cmT: Option<C1>,
     pub F: FC,              // F circuit
     pub x: Option<CF1<C1>>, // public inputs (u_{i+1}.x)
@@ -285,6 +286,7 @@ where
             u_i: None,
             U_i: None,
             U_i1: None,
+            r_nonnat: None,
             cmT: None,
             F: F_circuit,
             x: None,
@@ -340,6 +342,10 @@ where
         let U_i1 = CommittedInstanceVar::<C1>::new_witness(cs.clone(), || {
             Ok(self.U_i1.unwrap_or(u_dummy_native.clone()))
         })?;
+        let r_nonnat =
+            NonNativeFieldVar::<C1::BaseField, C1::ScalarField>::new_witness(cs.clone(), || {
+                Ok(self.r_nonnat.unwrap_or_else(CF2::<C1>::zero))
+            })?;
         let cmT =
             NonNativeAffineVar::new_witness(cs.clone(), || Ok(self.cmT.unwrap_or_else(C1::zero)))?;
         let x =
@@ -363,7 +369,6 @@ where
         let (u_i_x, U_i_vec) =
             U_i.clone()
                 .hash(&crh_params, i.clone(), z_0.clone(), z_i.clone())?;
-
         // check that h == u_i.x
         (u_i.x[0]).conditional_enforce_equal(&u_i_x, &is_not_basecase)?;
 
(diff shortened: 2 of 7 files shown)


folding-schemes/src/folding/nova/cyclefold.rs
```diff
@@ -330,24 +331,29 @@ where
...
+        // check that the points coordinates are placed as the public input x: x == [p1, p2, p3]
+        let r: FpVar<CF2<C>> = Boolean::le_bits_to_fp_var(&r_bits)?;
+        let points_coords: Vec<FpVar<CF2<C>>> = [
+            vec![r],
+            p1.clone().to_constraint_field()?,
+            p2.clone().to_constraint_field()?,
+            p3.clone().to_constraint_field()?,
+        ]
+        .concat();
+        points_coords.enforce_equal(&x)?;
```
arnaucube, line 349: This is the main change, the motivation for this PR solving #48

folding-schemes/src/folding/nova/cyclefold.rs
```diff
@@ -330,24 +331,29 @@ where
...
+        // check that the points coordinates are placed as the public input x: x == [r, p1, p2, p3]
+        let r: FpVar<CF2<C>> = Boolean::le_bits_to_fp_var(&r_bits)?;
+        let points_coords: Vec<FpVar<CF2<C>>> = [
+            vec![r],
+            p1.clone().to_constraint_field()?,
+            p2.clone().to_constraint_field()?,
+            p3.clone().to_constraint_field()?,
+        ]
+        .concat();
+        points_coords.enforce_equal(&x)?;
```
arnaucube, line 349: This is the main change, the motivation for this PR solving #48

folding-schemes/src/folding/nova/circuits.rs
```diff
@@ -391,6 +406,11 @@ where
         )?;
         let r = Boolean::le_bits_to_fp_var(&r_bits)?;
 
+        // enforce that the input r_nonnat as bits matches the in-circuit computed r_bits
+        let r_nonnat_bits: Vec<Boolean<C1::ScalarField>> =
+            r_nonnat.to_bits_le()?.into_iter().take(128).collect();
```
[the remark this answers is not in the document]
arnaucube: Oh this is a mistake from my side! this should come from the global constant `N_BITS_RO`. The reason for it being 128 comes from [Srinath Setty](https://microsoft.com/en-us/research/people/srinath/): *"In Nova, soundness error ≤ 2/|S|, where S is the subset of the field F from which the challenges are drawn. In this case, we keep the size of S close to 2^128."* (documented in the [constant](https://github.com/privacy-scaling-explorations/folding-schemes/blob/main/folding-schemes/src/constants.rs#L2)).
[Updated](https://github.com/privacy-scaling-explorations/folding-schemes/compare/872bdcc2706b2f57a0b62d13466c65b199c75d01..a401af0adf8a6e7b23292b7ee92fe838fc289cbb) this PR to use the constant instead of the hardcoded value.
Thanks for pointing it out!

folding-schemes/src/folding/nova/decider_eth_circuit.rs
```diff
@@ -562,10 +563,16 @@ where
...
-        let (cmE_x_limbs, cmE_y_limbs): (Vec<C::ScalarField>, Vec<C::ScalarField>) =
-            point_to_nonnative_limbs::<C>(U_i.cmE)?;
-        let (cmW_x_limbs, cmW_y_limbs): (Vec<C::ScalarField>, Vec<C::ScalarField>) =
-            point_to_nonnative_limbs::<C>(U_i.cmW)?;
+        let (cmE_x_limbs, cmE_y_limbs, _): (
+            Vec<C::ScalarField>,
+            Vec<C::ScalarField>,
+            Vec<C::ScalarField>,
+        ) = point_to_nonnative_limbs::<C>(U_i.cmE)?;
+        let (cmW_x_limbs, cmW_y_limbs, _): (
```
dmpierre, line 571: Is there a reason to ignoring the flag indicating if this is the infinity point when computing challenges?

arnaucube: Yes, because we're already identifying them as [0,1] when infinity point, so was ignoring the 'infinity' boolean. But changed mind regarding the whole 'infinity' bit/boolean, and I think that we can directly remove it from the point representation. Updated the PR to reflect that. This will also reduce the needed amount of constraints.

folding-schemes/src/folding/circuits/nonnative.rs
```diff
@@ -88,7 +114,12 @@ where
                 &C::BaseField::one(),
                 optimization_type,
             )?;
-        return Ok((x, y));
+        let infinity =
```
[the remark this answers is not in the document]
arnaucube: Regarding the question, they were as `NonNativeFieldVar` because when calling `get_cm_coordinates` it was returning 3 `NonNativeFieldVars` which can then be all together set into an array of `NonNativeFieldVar` (having one of the three as `FpVar` or `Boolean<CF>` would be a bit more tedious. But on the same time the nonnative representation was using extra constraints that were not really needed, and could be avoided as just a `Boolean<CF>` or `FpVar`.
But anyways, we don't actually need it (more details in the other comment: https://github.com/privacy-scaling-explorations/folding-schemes/pull/81#discussion_r1544671766), so just removed the infinity flag for the points representation.

folding-schemes/src/folding/nova/cyclefold.rs
```diff
@@ -27,8 +27,8 @@ use super::CommittedInstance;
 use crate::constants::N_BITS_RO;
 use crate::Error;
 
-// publi inputs length for the CycleFoldCircuit, |[p1.x,y, p2.x,y, p3.x,y]|
-pub const CF_IO_LEN: usize = 6;
+// publi inputs length for the CycleFoldCircuit: |[r, p1.x,y, p2.x,y, p3.x,y]|
```
dmpierre, line 30: spotted small typo
```suggestion
// public inputs length for the CycleFoldCircuit: |[r, p1.x,y, p2.x,y, p3.x,y]|
```

dmpierre (approved):
lgtm, changes are indeed minimal without adding constraints required from the infinity point! spotted a small typo while looking over the changes.

folding-schemes/src/folding/nova/cyclefold.rs
arnaucube: oh right, thanks, amended

han0110 (approved):
LGTM!

PR #84, opened 2024-04-04
Add the digest of the Relaxed R1CS instance for CycleFold as a public input to `AugmentedFCircuit`

### LINKED ISSUE #83: Should `AugmentedFCircuit` for Nova include an additional public input for CycleFold?
In `AugmentedFCircuit` for Nova, we currently check:
https://github.com/privacy-scaling-explorations/sonobe/blob/fe9a488f63fd8a131119cd3e40ec4d80dcf4690a/folding-schemes/src/folding/nova/circuits.rs#L486-L503
1. `cf1_U_i1` is correctly folded from `cf_U_i` and `cf1_u_i`, and
2. `cf_U_i1` is correctly folded from `cf1_U_i1` and `cf2_u_i`.

However, these checks are insufficient to ensure that `cf_U_i1` is indeed the result from the CycleFold circuit, as `cf_U_i1` is a witness rather than a public input.

Being curious about this, I looked into [arecibo](https://github.com/lurk-lab/arecibo)'s implementation, where the correctness of folding `U_c` (our `cf_U_i1`) is also checked in the [circuit](https://github.com/lurk-lab/arecibo/blob/5268c2084821955c5953c29c7e179b2052523b5d/src/cyclefold/nova_circuit.rs#L372-L379), but the circuit further computes `hash_c`, [the digest of `U_c`](https://github.com/lurk-lab/arecibo/blob/5268c2084821955c5953c29c7e179b2052523b5d/src/cyclefold/nova_circuit.rs#L500-L510) (`Unew_c` in the referred code is equal to `U_c` in non-base cases), which is [treated the second public input to the circuit](https://github.com/lurk-lab/arecibo/blob/5268c2084821955c5953c29c7e179b2052523b5d/src/cyclefold/nova_circuit.rs#L513). I think this makes more sense as it allows one to verify if the in-circuit variable `cf_U_i1` is indeed the desired one.

Btw, [this hackmd](https://hackmd.io/SBvAur_2RQmaduDi7gYbhw) explains arecibo's implementation in detail, and it also mentions that "The Nova augmented circuit has a public IO vector x of length 2".

Therefore, I guess we should also add another public input to `AugmentedFCircuit` for full soundness, as well as the logic for computing `cf_U_i1`'s digest inside and outside the circuit.

Please correct me if I misunderstood something :)

winderica:
Closes #83. Now the augmented circuit additionally checks:
1.  $u_i.x[1] = H(U^{\mathsf{cf}}_i)$
2. $u_{i+1}.x[1] = H(U^{\mathsf{cf}}_{i+1})$

### DIFF
--- a/folding-schemes/src/folding/nova/cyclefold.rs
+++ b/folding-schemes/src/folding/nova/cyclefold.rs
@@ -109,7 +109,7 @@ where
             nonnative_field_var_to_constraint_field(&self.u)?,
             self.x
                 .iter()
-                .map(|i| nonnative_field_var_to_constraint_field(i))
+                .map(nonnative_field_var_to_constraint_field)
                 .collect::<Result<Vec<_>, _>>()?
                 .concat(),
             cmE_elems,


folding-schemes/src/folding/nova/cyclefold.rs
```diff
@@ -82,6 +86,87 @@ where
     }
 }
 
+// A more efficient version of `NonNativeFieldVar::to_constraint_field`
+fn nonnative_field_var_to_constraint_field<TargetField: PrimeField, BaseField: PrimeField>(
```
arnaucube, line 90: This is great! How would you see to put this method in the `folding-schemes/src/folding/circuits/*` dir? Because probably we're going to use it in the other folding schemes.

folding-schemes/src/folding/nova/mod.rs
```diff
@@ -93,6 +97,84 @@ where
     }
 }
 
+// The out-circuit counterpart of `nonnative_field_var_to_constraint_field` in `nova/cyclefold.rs`
+fn nonnative_field_to_field_elements<TargetField: PrimeField, BaseField: PrimeField>(
```
arnaucube, line 101: As in the prev comment, how would you see to move this method in the `folding-schemes/src/folding/circuits/*` dir?

arnaucube (commented):
This is awesome! Thank you very much! :raised_hands:
You improved many things in the code, not just the missing link between the hash and the instance public input.
Just left a small suggestion for to move a couple of methods.
But again, this is great, thank you very much!

arnaucube (approved):
LGTM! Thank you very much for this PR! :raised_hands:

folding-schemes/src/folding/nova/circuits.rs
```diff
@@ -379,7 +381,7 @@ where
         )?;
         let zero_y = NonNativeFieldVar::<C1::BaseField, C1::ScalarField>::new_constant(
             cs.clone(),
-            C1::BaseField::one(),
+            C1::BaseField::zero(),
         )?;
         (u_i.cmE.x.is_eq(&zero_x)?).conditional_enforce_equal(&Boolean::TRUE, &is_not_basecase)?;
         (u_i.cmE.y.is_eq(&zero_y)?).conditional_enforce_equal(&Boolean::TRUE, &is_not_basecase)?;
```
arnaucube, line 387: small nit: maybe `zero_y` can now be removed and just use a single var containing `zero` for both `cmE.x` and `cmE.y`?

folding-schemes/src/folding/nova/decider_eth_circuit.rs
```diff
@@ -419,13 +419,13 @@ where
         )?;
         let zero_y = NonNativeFieldVar::<C1::BaseField, C1::ScalarField>::new_constant(
             cs.clone(),
-            C1::BaseField::one(),
+            C1::BaseField::zero(),
         )?;
         u_i.cmE.x.enforce_equal(&zero_x)?;
         u_i.cmE.y.enforce_equal(&zero_y)?;
```
arnaucube, line 425: same as in [the other `zero_y` comment](https://github.com/privacy-scaling-explorations/sonobe/pull/84/files#r1559713822), maybe now can just use a single var for `zero` for `cmE.x` & `cmE.y`?

folding-schemes/src/folding/nova/cyclefold.rs
```diff
@@ -82,6 +87,60 @@ where
...
+        let cmE_is_inf = cmE_elems.pop().unwrap();
+        let cmW_is_inf = cmW_elems.pop().unwrap();
+        // Concatenate `cmE_is_inf` and `cmW_is_inf` to save constraints for CRHGadget::evaluate
+        let is_inf = cmE_is_inf.double()? + cmW_is_inf;
+
+        Ok([
+            nonnative_field_var_to_constraint_field(&self.u)?,
+            self.x
+                .iter()
+                .flat_map(|i| nonnative_field_var_to_constraint_field(i).unwrap())
```
dmpierre, line 112: I think gracefully handling the `SynthesisError`, if any, might be better here no?

dmpierre (commented):
Hi, thanks a lot for this! I have a single comment to make, notably about an `unwrap()` call.

dmpierre (approved):

PR #86, opened 2024-04-06
Reduce the number of constraints in `AugmentedFCircuit` for Nova

winderica:
This PR depends on #84.

For the test `folding::nova::tests::test_ivc`:
Before 520db300f3f8929b211b91897ec1ed4387d06d7a: 138240
After 520db300f3f8929b211b91897ec1ed4387d06d7a: 86756 (1.6x improvement)

Two notable optimization techniques are employed:
1. Instead of allocating two witness variables `a, b` and enforce their equality by calling `a.conditional_enforce_equal(&b, &cond)`, we can avoid the allocation of `b` and directly set `b = a`. The former might be costly due to the checks in allocation and `conditional_enforce_equal`. See `nova/circuits.rs` for details.
2. Before 520db300f3f8929b211b91897ec1ed4387d06d7a, `NonNativeFieldVar::to_constraint_field` was majorly called for generating the inputs (preimage) to hash functions. However, it turns out that the underlying conversion strategy (optimized for weight), although aimed to reducing the number of limbs, is suboptimal for reducing the length of hash preimage. We can go further by maximizing the number of bits per limb, thereby minimizing the preimage length. See `circuits/nonnative.rs` for details.

### DIFF
--- a/folding-schemes/src/folding/nova/circuits.rs
+++ b/folding-schemes/src/folding/nova/circuits.rs
@@ -17,11 +17,10 @@ use ark_r1cs_std::{
     fields::{fp::FpVar, nonnative::NonNativeFieldVar, FieldVar},
     groups::GroupOpsBounds,
     prelude::CurveVar,
-    ToBitsGadget, ToConstraintFieldGadget,
+    R1CSVar, ToConstraintFieldGadget,
 };
 use ark_relations::r1cs::{ConstraintSynthesizer, ConstraintSystemRef, Namespace, SynthesisError};
-use ark_std::fmt::Debug;
-use ark_std::Zero;
+use ark_std::{fmt::Debug, Zero};
 use core::{borrow::Borrow, marker::PhantomData};
 
 use super::{
@@ -30,9 +29,9 @@ use super::{
     },
     CommittedInstance,
 };
-use crate::constants::N_BITS_RO;
-use crate::folding::circuits::nonnative::{point_to_nonnative_limbs, NonNativeAffineVar};
+use crate::folding::circuits::nonnative::{nonnative_affine_to_field_elements, NonNativeAffineVar};
 use crate::frontend::FCircuit;
+use crate::{constants::N_BITS_RO, folding::circuits::nonnative::nonnative_field_var_from_le_bits};
 
 /// CF1 represents the ConstraintField used for the main Nova circuit which is over E1::Fr, where
 /// E1 is the main curve where we do the folding.
@@ -106,10 +105,8 @@ where
         let U_vec = [
             vec![self.u],
             self.x,
-            self.cmE.x.to_constraint_field()?,
-            self.cmE.y.to_constraint_field()?,
-            self.cmW.x.to_constraint_field()?,
-            self.cmW.y.to_constraint_field()?,
+            self.cmE.to_constraint_field()?,
+            self.cmW.to_constraint_field()?,
         ]
         .concat();
         let input = [vec![i], z_0, z_i, U_vec.clone()].concat();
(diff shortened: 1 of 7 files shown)


folding-schemes/src/folding/nova/cyclefold.rs
```diff
@@ -189,13 +182,36 @@ pub struct NIFSFullGadget<C: CurveGroup, GC: CurveVar<C, CF2<C>>> {
...
     for<'a> &'a GC: GroupOpsBounds<'a, C, GC>,
 {
+    pub fn fold_committed_instance(
+        // assumes that r_bits is equal to r_nonnat just that in a different format
+        r_bits: Vec<Boolean<CF2<C>>>,
+        r_nonnat: NonNativeFieldVar<C::ScalarField, CF2<C>>,
+        cmT: GC,
+        // ci1 is assumed to be always with cmE=0, u=1 (checks done previous to this method)
+        ci1: CycleFoldCommittedInstanceVar<C, GC>,
+        ci2: CycleFoldCommittedInstanceVar<C, GC>,
```
arnaucube, line 200: This was a mistake made by me in a previous commit, where this comment should be for `ci2`, which is the one which is assumed to be with `cmE=0` and `u=1`, right? Which comes from `u_i`.
The code of the function already uses it correctly since does
`cmE: cmT.scalar_mul_le(r_bits.iter())? + ci1.cmE`
where `ci2.cmE` is skipped since is `0`. So it would just be updating the comment to `ci2` instead of `ci1`.
(same for the `verify` method, for the comment at line 220)

folding-schemes/src/folding/nova/circuits.rs
```diff
@@ -389,60 +401,38 @@ where
...
         )?;
-
-        u_i1_x.enforce_equal(&x)?;
+        let (u_i1_x_base, _) = CommittedInstanceVar::new_constant(cs.clone(), u_dummy)?.hash(
+            &crh_params,
+            FpVar::<CF1<C1>>::one(),
+            z_0.clone(),
+            z_i1.clone(),
+        )?;
+        let x = FpVar::new_input(cs.clone(), || Ok(self.x.unwrap_or(u_i1_x_base.value()?)))?;
```
arnaucube, line 430: Should the `unwrap_or(u_i1_x_base.value())` default to `zero` instead of `u_i1_x_base.value()`?
Because with the current approach, when `self.x` is `None` the next check is checking if `x == u_i1_x_base`, where `x` is loaded directly from `u_i1_x_base`.

folding-schemes/src/folding/nova/circuits.rs
```diff
@@ -451,78 +441,88 @@ where
...
+        let (cf_u_i1_x, _) = cf_U_i1.clone().hash(&crh_params)?;
+        let (cf_u_i1_x_base, _) =
+            CycleFoldCommittedInstanceVar::new_constant(cs.clone(), cf_u_dummy)?
+                .hash(&crh_params)?;
+        let cf_x = FpVar::new_input(cs.clone(), || {
+            Ok(self.cf_x.unwrap_or(cf_u_i1_x_base.value()?))
+        })?;
+        cf_x.enforce_equal(&is_basecase.select(&cf_u_i1_x_base, &cf_u_i1_x)?)?;
+
+        println!("Constraints: {}", cs.num_constraints());
```
arnaucube, line 525: Maybe this println can be removed

arnaucube (commented):
This is great! Great improvement! :raised_hands:
Just left some small nits

folding-schemes/src/folding/nova/circuits.rs
winderica: IMHO we can pass an arbitrary value to the variable allocation function `new_{witness, input}` without affecting the circuit's soundness? What I am doing here is just creating fallback values that are always satisfiable when the inputs are `None` (e.g., those used in `Nova::init`). In `Nova::prove_step`, all the inputs are correctly assigned, and thus the `unwrap` call never returns the default value.

Let me also elaborate on my first claim above.

We build circuits in an input-agnostic way: although the circuit should accept correct inputs and reject incorrect ones, we are unaware if the user will assign correct or incorrect inputs to the circuit.

However, it would be inconvenient if the circuit developer needs to separate the circuit implementation and input assignment, since the developer needs to figure out which input corresponds to which gate. I guess that's why circuit-building libraries allow us to specify the value of an input. But this does not mean the value we specified in `new_{witness, input}` is what the user supplies. For a valid proof, the user has to provide correct values, even if we pass incorrect values to `new_{witness, input}`. Vice versa, the user can provide incorrect values to generate an invalid proof, even if we always pass the correct ones.

I hope this makes sense, though it is also fine for me to change it to `unwrap_or_default()` 😄

arnaucube: Right, I see, makes total sense. Thanks for the explanation! ^^

arnaucube (approved):
Thanks a lot for these improvements!

dmpierre (approved):
Hi, thanks for this work as well; this is a neat thing to bypass how arkworks builds limbs for non native vars.
lgtm!

PR #90, opened 2024-04-23
Full flow example

arnaucube:
- adds full flow example (from folding to verifying in the EVM)
- moves all the examples into a single root dir `examples`


han0110 (approved):
LGTM!

PR #91, opened 2024-04-29
Circom external inputs

arnaucube:
- Updates the `FoldingScheme` trait (and `nova` impl) to accept `external_inputs` at each `.prove_step` call (instead of setting them at the initialization call)
- Add external inputs to Circom's frontend
- move circom R1CS file read at initializer instead of each ivc step
- add `examples/circom_full_flow.rs` containing an example that goes from setting a circom circuit and folding it to verifying the proof in the EVM (like the `full_flow.rs` but instead of arkworks frontend using circom frontend). To showcase how to use the lib in Circom's case.

### DIFF
--- a/folding-schemes/src/frontend/mod.rs
+++ b/folding-schemes/src/frontend/mod.rs
@@ -28,6 +28,7 @@ pub trait FCircuit<F: PrimeField>: Clone + Debug {
         &self,
         i: usize,
         z_i: Vec<F>,
+        external_inputs: Vec<F>, // inputs that are not part of the state
     ) -> Result<Vec<F>, Error>;
 
     /// generates the constraints for the step of F for the given z_i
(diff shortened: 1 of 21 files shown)


folding-schemes/src/frontend/mod.rs
```diff
@@ -14,12 +14,16 @@ pub trait FCircuit<F: PrimeField>: Clone + Debug {
     type Params: Debug;
 
     /// returns a new FCircuit instance
-    fn new(params: Self::Params) -> Self;
+    fn new(params: Self::Params) -> Result<Self, Error>;
```
dmpierre, line 17: Out of curiosity why turning this into a `Result` now?

arnaucube: Because there are cases where an error might be needed to be returned due some internal operation at the `new` call. For example at the circom frontend: https://github.com/privacy-scaling-explorations/sonobe/blob/3aa4483962bacb4a530414e9136915012437af4f/folding-schemes/src/frontend/circom/mod.rs#L34
Before we were not needing that specifically in the circom frontend because the r1cs reading was being done at each folding step, but this was extra work that we could avoid, so now it is done only once at the beginning instead of at each fold, and it can return an error thus needs the `Result`.

PR #101, opened 2024-05-23
Stabilize circom frontend

dmpierre:
This PR aims at stabilizing sonobe's circom frontend. To do this we:
1. Fixed the allocation of our IVC inputs when using `circom-compat`. A PR on a `circom-compat` fork has been opened [here](https://github.com/arnaucube/circom-compat/pull/1) to choose how our IVC inputs are allocated (as witnesses and public inputs) and to adjust circom's R1CS indexes when wrapped by our `AugmentedFCircuit`.
2. Disabled default reading of circom's wire mapping when loading R1CS files - as what is being done by [`circom-compat`](https://github.com/arkworks-rs/circom-compat/blob/170b10fc9ed182b5f72ecf379033dda023d0bf07/src/circom/builder.rs#L64).
3. Added a test using a previously non-working circuit; the R1CS of keccak.


folding-schemes/src/frontend/circom/test_folder/keccak256-circom/circuits/keccak.circom
```diff
@@ -0,0 +1,186 @@
+// Keccak256 hash function (ethereum version).
+// For LICENSE check https://github.com/vocdoni/keccak256-circom/blob/master/LICENSE
```
arnaucube, line 2: Maybe we could use a simpler circuit that does not require pasting here the entire circuits of the https://github.com/vocdoni/keccak256-circom repo?

Ideally not even importing external projects, but just a self-contained & simpler circuit with few lines of code that fails with the version of the frontend/circom before this PR.

For example, in the `external_inputs.circom`, by using a template like `IsZero`:
```circom
pragma circom 2.0.3;

// This template is from circomlib/circuits/comparators.circom
template IsZero() {
    signal input in;
    signal output out;

    signal inv;

    inv <-- in!=0 ? 1/in : 0;

    out <== -in*inv +1;
    in*out === 0;
}


/*
    z_{i+1} == z_i^3 + z_i * external_input[0] + external_input[1]
*/
template Example () {
    signal input ivc_input[1]; // IVC state
    signal input external_inputs[2]; // not state

    signal output ivc_output[1]; // next IVC state

    component check_input = IsZero(); // this is what triggers the error prior to this PR
    check_input.in <== ivc_input[0];
    check_input.out === 0;

    signal temp1;
    signal temp2;
    
    temp1 <== ivc_input[0] * ivc_input[0];
    temp2 <== ivc_input[0] * external_inputs[0];
    ivc_output[0] <== temp1 * ivc_input[0] + temp2 + external_inputs[1];
}

component main {public [ivc_input]} = Example();
```
this will already fail with the version prior to this PR and could be used to test the fixes from this PR on the Circom frontend. Maybe with some other edge cases that were triggering the error prior to the PR.

arnaucube (changes requested):
This is great!!! Great job! :raised_hands:

Tomorrow I'll try to use it in some other external circom projects, but it looks great so far!

folding-schemes/src/frontend/circom/test_folder/keccak256-circom/circuits/keccak.circom
dmpierre: Yes, the keccak circuit, though covers everything we need, is not very practical to test. I replaced it with more lightweight circuits. 
We still depend on one copied-pasted `circomlib` circuit (your suggested`IsZero` template): the test will re-use it for two different cases, namely with and without external inputs. Tests will also check that circom defined constraints are well enforced by feeding in erroneous inputs.

examples/circom_full_flow.rs
```diff
@@ -57,9 +57,9 @@ fn main() {
 
     // initialize the Circom circuit
     let r1cs_path =
-        PathBuf::from("./folding-schemes/src/frontend/circom/test_folder/external_inputs.r1cs");
+        PathBuf::from("./folding-schemes/src/frontend/circom/test_folder/full_flow_example.r1cs");
     let wasm_path = PathBuf::from(
-        "./folding-schemes/src/frontend/circom/test_folder/external_inputs_js/external_inputs.wasm",
+        "./folding-schemes/src/frontend/circom/test_folder/full_flow_example_js/full_flow_example.wasm",
```
arnaucube, line 62: I was thinking if this example could not reuse the already existing `folding-schemes/src/frontend/circom/test_folder/with_external_inputs.circom` circuit instead of duplicating it into `full_flow_example.circom`, since the circuits are identical exept for the 'IsZero' template usage.
With the idea to do it similarly to what was done before this PR, where the `circom_full_flow.rs` example reused the circom test circuits.

But when I've tried the `example/circom_full_flow.rs` example using the `with_external_inputs.circom` circuit it fails with `SynthesisError(AssignmentMissing)`.

To reproduce it just change these couple of lines to load the `with_external_inputs.{r1cs/wasm}` instead of `full_flow_example.{r1cs/wasm}` and run the `examples/circom_full_flow.rs` example.

dmpierre: Weird, I don't get `Err(Assignment)` missing but `Err(Unsatisfiable)` which is expected - this is related to that [issue](https://github.com/privacy-scaling-explorations/sonobe/issues/104) and is why I switched to a circuit that can be satisfied with a zero vector assigned.

arnaucube: You're right! My bad, I was using somehow a wrong version of the branch!

arnaucube: Now that the sanity-check does not trigger the error (since the commit https://github.com/privacy-scaling-explorations/sonobe/pull/101/commits/c278b931e96af480292407d38626ff40d55d9aa9 ), how would you see to update this to reuse the `with_external_inputs.circom` related files and removing the (almost) duplicate `full_flow_example.circom`?

dmpierre: Yup, good suggestion. Done here [bdf6578](https://github.com/privacy-scaling-explorations/sonobe/pull/101/commits/bdf657801ee6415c78e7a320f3a9588cedb36367)

.github/workflows/typos.toml
```diff
@@ -1,2 +1,6 @@
 [default.extend-words]
 groth = "groth"
+
+[type.po]
+extend-glob = ["*.circom"]
+check-file = false
```
arnaucube, line 6: small nit: do we need this? I understand that this was added to prevent lints over the imported vocdoni/keccak256-circom repo files, but I assume that no longer is needed, and it's good to be able to catch typos in the circom files.

arnaucube (approved):
This is great!! Great work! :rocket: 
Big milestone fixing it! Now we can fully *fodl* Circom circuits ^^

PR #106, opened 2024-05-29
BN254 and Grumpkin Poseidon configurations

### LINKED ISSUE #85: Poseidon configuration params for Ethereum use case
Currently the tests that require usage of Poseidon are using the [`poseidon_test_config()`](https://github.com/privacy-scaling-explorations/sonobe/blob/b8db622a08ff7c7d11a0003f8bd9a3a8955649e8/folding-schemes/src/transcript/poseidon.rs#L119) params generator.
We need to have a specific method that offers the `poseidon_config` for the Ethereum use case (over Bn254's scalar field), to be used in non-test use cases that want to use the Ethereum Decider to verify the proofs. Ideally using the same config as in [circomlib](https://github.com/iden3/circomlib/blob/master/circuits/poseidon.circom), [noir-stdlib](https://github.com/noir-lang/noir/pull/768), etc.
And maybe the same for [pasta curves](https://o1-labs.github.io/proof-systems/specs/pasta.html) ([pallas](https://github.com/arkworks-rs/algebra/tree/master/curves/pallas) and [vesta](https://github.com/arkworks-rs/algebra/tree/master/curves/vesta)) scalar fields.

ax0:
This PR resolves #85 by including Poseidon configurations for the scalar fields of the BN254 and Grumpkin curves. The former configuration agrees with Circom's Poseidon(4) while the latter was generated using the reference implementation script. For all other fields, the configuration falls back to the test configuration used previously. Tests have been modified to use this curve cycle.


arnaucube (approved):
Great!! ^^

folding-schemes/src/transcript/poseidon/grumpkin.rs
```diff
@@ -0,0 +1,518 @@
+use crate::utils::vec::mat_from_str_mat;
+use ark_crypto_primitives::sponge::poseidon::PoseidonConfig;
+use ark_ff::{BigInteger, PrimeField};
+use ark_grumpkin::Fr;
+
+// Poseidon constants for Grumpkin curve as generated by
+// https://extgit.iaik.tugraz.at/krypto/hadeshash/-/blob/master/code/generate_params_poseidon.sage
```
dmpierre, line 7: Proposing to also add sage script's arguments to ease reproducibility. Checked on my side and correct me if I'm wrong, but should be those ones?

```suggestion
// https://extgit.iaik.tugraz.at/krypto/hadeshash/-/blob/master/code/generate_params_poseidon.sage
// generated with `sage generate_params_poseidon.sage 1 0 254 5 5 128 0x30644e72e1
31a029b85045b68181585d97816a916871ca8d3c208c16d87cfd47`
```

dmpierre (changes requested):
Thanks for this!
Proposing to add a comment on the arguments used with sage when generating parameters.

dmpierre (approved):

PR #107, opened 2024-05-31
Add serde capabilites to `Nova`

dmpierre:
This PR aims to add serialization and deserialization capabilities to `Nova`.

For serialization, we are able to derive the `CanonicalSerialize` trait for most of the fields of `struct Nova`. Two notable exceptions are:
1. `KZG::ProverKey`, implemented with `impl<'a, C: CurveGroup> CanonicalSerialize for ProverKey<'a, C>`. This is required since `powers_of_g` are wrapped within a `Cow`.
2. `PoseidonParams`, since we can not implement an external trait to an external struct - Both `PoseidonParams` and `PoseidonParams` are imported from `arkworks`.

We could not implement the [`CanonicalDeserialize`](https://docs.rs/ark-serialize/latest/ark_serialize/trait.CanonicalDeserialize.html) trait on `Nova`. `CanonicalDeserialize` requires the types to be deserialized to also satisfy the [`Valid`](https://docs.rs/ark-serialize/latest/ark_serialize/trait.Valid.html) trait, itself requiring types to satisfy `Sync + Sized`. Since, [`CurveVar`](https://github.com/arkworks-rs/r1cs-std/blob/4020fbc22625621baa8125ede87abaeac3c1ca26/src/groups/mod.rs#L32) does not implement `Sync`, we had to resort to not implement the trait itself.


folding-schemes/src/folding/nova/mod.rs
```diff
@@ -854,6 +1039,77 @@ pub mod tests {
...
+
+        assert_eq!(nova.i, deserialized_nova.i);
+
+        let num_steps: usize = 3;
+        for _ in 0..num_steps {
+            deserialized_nova.prove_step(vec![]).unwrap();
+            nova.prove_step(vec![]).unwrap();
+        }
+
+        assert_eq!(deserialized_nova.w_i, nova.w_i);
```
dmpierre, line 1110: We show here that we end up with the same witness, both for the pre-serialized and post-deserialized structs.

folding-schemes/src/folding/nova/mod.rs
```diff
@@ -249,6 +250,187 @@ where
...
+        FC: FCircuit<C1::ScalarField, Params = ()>,
+        CS1: CommitmentScheme<C1>,
+        CS2: CommitmentScheme<C2>,
+    > Nova<C1, GC1, C2, GC2, FC, CS1, CS2>
+where
+    CS1::ProverParams: CanonicalDeserialize,
+    CS2::ProverParams: CanonicalDeserialize,
+{
+    #[allow(dead_code)]
+    fn deserialize_nova<R: std::io::prelude::Read>(
```
dmpierre, line 369: We can't derive or implement the `CanonicalDeserialize` trait since `CurveVar` notably does not implement the `Sync` trait, which is required.

folding-schemes/src/folding/nova/mod.rs
```diff
@@ -249,6 +250,187 @@ where
...
+        self.z_i.serialize_with_mode(&mut writer, compress)?;
+        self.w_i.serialize_with_mode(&mut writer, compress)?;
+        self.u_i.serialize_with_mode(&mut writer, compress)?;
+        self.W_i.serialize_with_mode(&mut writer, compress)?;
+        self.U_i.serialize_with_mode(&mut writer, compress)?;
+        self.cf_W_i.serialize_with_mode(&mut writer, compress)?;
+        self.cf_U_i.serialize_with_mode(&mut writer, compress)?;
+        self.r1cs.serialize_with_mode(&mut writer, compress)?;
+        self.cf_r1cs.serialize_with_mode(&mut writer, compress)?;
+        self.poseidon_config
```
dmpierre, line 283: We serialize `PoseidonParams` since we can't implement an external trait (`CanonicalDeserialize`) on an external type (`PoseidonParams`).

arnaucube (changes requested):
Nice!

folding-schemes/src/folding/nova/mod.rs
arnaucube: Same as in https://github.com/privacy-scaling-explorations/sonobe/pull/107#discussion_r1627145921

.github/workflows/ci.yml
```diff
@@ -60,7 +60,9 @@ jobs:
           curl -sSfL https://github.com/ethereum/solidity/releases/download/v0.8.4/solc-static-linux -o /usr/local/bin/solc
           chmod +x /usr/local/bin/solc
       - name: Execute compile.sh to generate .r1cs and .wasm from .circom
-        run: bash ./folding-schemes/src/frontend/circom/test_folder/compile.sh
+        run: ./folding-schemes/src/frontend/circom/test_folder/compile.sh
+      - name: Install clang
```
dmpierre, line 64: I updated a few steps so that it runs locally

folding-schemes/src/folding/nova/mod.rs
```diff
@@ -857,6 +862,77 @@ pub mod tests {
         );
     }
 
+    #[test]
+    fn test_serde_nova() {
```
arnaucube, line 866: should this test be under `nova/serialize.rs` since is testing the methods from that file?

folding-schemes/src/folding/nova/serialize.rs
```diff
@@ -0,0 +1,216 @@
...
+    C1: CurveGroup<BaseField = C2::ScalarField, ScalarField = C2::BaseField>,
+    for<'a> &'a GC1: GroupOpsBounds<'a, C1, GC1>,
+    for<'a> &'a GC2: GroupOpsBounds<'a, C2, GC2>,
+    GC1: CurveVar<C1, <C2 as Group>::ScalarField>,
+    GC1: ToConstraintFieldGadget<<C2 as Group>::ScalarField>,
+    GC2: CurveVar<C2, <C2 as CurveGroup>::BaseField>,
+    CS1::ProverParams: CanonicalDeserialize,
+    CS2::ProverParams: CanonicalDeserialize,
+{
+    #[allow(dead_code)]
```
arnaucube, line 152: seems that this `#[allow(dead_code)]` is not needed?

dmpierre: Had `cargo clippy` complaining in earlier commits, good catch!

folding-schemes/src/folding/nova/serialize.rs
```diff
@@ -0,0 +1,306 @@
...
+            + self.poseidon_config.full_rounds.serialized_size(compress)
+            + self
+                .poseidon_config
+                .partial_rounds
+                .serialized_size(compress)
+            + self.poseidon_config.alpha.serialized_size(compress)
+            + self.poseidon_config.ark.serialized_size(compress)
+            + self.poseidon_config.mds.serialized_size(compress)
+            + self.poseidon_config.rate.serialized_size(compress)
+            + self.poseidon_config.capacity.serialized_size(compress)
```
arnaucube, line 104: Sorry for the late comment :pray: , thought about it initially and discarded it, but now I'm giving it a second thought due the difference that it makes.

Do those params need to be serialized? The two R1CS, the Poseidon params, and the CommitmentScheme params.

They are quite big, and they can be regenerated without reading them from the serialized file. Also, they depend anyways on the definition of the configuration of the Nova struct (eg. even if importing the r1cs data, the Nova struct requires to have defined the FCircuit and internally will compute the r1cs for the AugmentedFCircuit).

What I'm thinking is that the
```
self.cs_params
self.cf_cs_params
self.r1cs
self.cf_r1cs
self.poseidon_config.full_rounds
self.poseidon_config.partial_rounds
self.poseidon_config.alpha
self.poseidon_config.ark
self.poseidon_config.mds
self.poseidon_config.rate
self.poseidon_config.capacity
```
parameters don't need to be serialized/deserialized, can be removed from the serialization, and can just be computed/taken as input in the deserialize method.


I've tested it locally, and the serialized data size went from `50 MB` to `6.4 MB` (`7.8x` reduction).

dmpierre: I thought that it would be more user friendly this way. But indeed, this is almost a 10x reduction, which can be massive when using large circuits. I kept deriving the `Canonical(De)Serialize` traits on commitment schemes and `R1CS` since I think it might still be useful for users.

[d29bf0f](https://github.com/privacy-scaling-explorations/sonobe/pull/107/commits/d29bf0f493159bde40f6c09cbcd8b3435dc96c32)

arnaucube: Nice, how would you see computing the r1cs data in the deserialize method too, avoiding the need for the user to obtain them outside?
The reasoning is that in the current test it's not a problem but because they already exist in the `nova` struct before serializing/deserializing, but in a real usage the dev deserializing back to Nova would not have at hand the r1cs data, and it's data that the lib can already obtain from the given FCircuit. 

It would be adding the lines from 300 to 315 from [nova/mod.rs#:300](https://github.com/privacy-scaling-explorations/sonobe/blob/5ea55cf54e1d1fc8ca523cab07f1c42fc6bade3f/folding-schemes/src/folding/nova/mod.rs#L300) into the `deserialize_nova` method and removing the inputs `r1cs` & `cf_r1cs`.

dmpierre: Yup good idea, wil make things easier [73ce876](https://github.com/privacy-scaling-explorations/sonobe/pull/107/commits/73ce87634c17b608f5da099c776d3856de9cb45c)

PR #110, opened 2024-06-05
Optimize native nimfs

arnaucube:
When we implemented the HyperNova multifolding (https://github.com/privacy-scaling-explorations/multifolding-poc) the focus was on understanding the protocol and the data structures, but not on making things run fast. When using real-world sized circuits, the hypernova methods implementations were taking too much computation time, since for example iterating over the full Boolean hypercube of size 16 takes too much time (>10min) and RAM.

This PR reimplements most of the HyperNova internal methods changing their approaches and reducing their computation times `>100x` depending on the method and the size.
(And the `>100x` is for matrices of size $2^9 \times 2^9$, where the real-world use case is of $2^{16} \times 2^{16}$, where the difference would be much more higher, but don't have enough RAM to compute the old version with those sizes to have concrete numbers. Thus the estimation is that probably the reduction for the real-world use case is about more than `1000x`).

| method                | matrix size   | old version seconds | new version seconds |
| --------------------- | ------------- | ------------------- | ------------------- |
| compute_g             | 2^8 x 2^8     | 16.48               | 0.16                |
| compute_g             | 2^9 x 2^9     | 122.62              | 0.51                |
| compute_Ls            | 2^8 x 2^8     | 9.73                | 0.11                |
| compute_Ls            | 2^9 x 2^9     | 67.16               | 0.38                |
| to_lcccs              | 2^8 x 2^8     | 4.56                | 0.21                |
| to_lcccs              | 2^9 x 2^9     | 67.65               | 0.84                |
| compute_sigmas_thetas | 2^8 x 2^8     | 12.86               | 0.13                |
| compute_sigmas_thetas | 2^9 x 2^9     | 100.01              | 0.51                |
| compute_Q             | 2^8 x 2^8     | 4.49                | 0.07                |
| compute_Q             | 2^9 x 2^9     | 70.77               | 0.55                |

### DIFF
--- a/folding-schemes/src/ccs/mod.rs
+++ b/folding-schemes/src/ccs/mod.rs
@@ -1,7 +1,7 @@
 use ark_ff::PrimeField;
 use ark_std::log2;
 
-use crate::utils::vec::*;
+use crate::utils::vec::{hadamard, mat_vec_mul, vec_add, vec_scalar_mul, SparseMatrix};
 use crate::Error;
 
 pub mod r1cs;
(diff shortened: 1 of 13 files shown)


dmpierre (approved):
Solid work with nice performance improvements!

PR #114, opened 2024-06-21
Generate poseidon constants

kilic:
Poseidon configuration in circom and arkworks/crypto-primitives already matches so it might not be necessary to hardcode constants. If hardcoding constants has another purpose such as performance concerns then this PR should be closed.

* Poseidon configuration of PR also targets 120 bit security
* Only `t = 5` or `rate = 4` is available as before
* First commit in this PR shows the equivalence of hardcoded and generated constants.


arnaucube (approved):
Oh this is cool! Thank you very much! ^^

dmpierre (approved):

CPerezz (approved):
Thanks so much for this @kilic <3

PR #129, opened 2024-07-16
Enable hiding commitments in nova and hypernova

dmpierre:
This PR aims to make it possible to specify blinding commitments. It consists in changing how generics are specified for the different structs used with nova and hypernova. We test our changes for [pedersen commitments](https://github.com/privacy-scaling-explorations/sonobe/blob/3099befc6a7c7d9a3a35b083cacfb0fa393b2cd9/folding-schemes/src/folding/nova/mod.rs#L939), since blinding factors have been [removed](https://github.com/privacy-scaling-explorations/sonobe/blob/3099befc6a7c7d9a3a35b083cacfb0fa393b2cd9/folding-schemes/src/commitment/kzg.rs#L130) from the KZG `commit()` method. 

This PR precedes #127, which adds the ability to generate zk IVC proofs for nova.

### DIFF
--- a/folding-schemes/src/folding/nova/mod.rs
+++ b/folding-schemes/src/folding/nova/mod.rs
@@ -166,16 +166,20 @@ impl<C: CurveGroup> Witness<C>
 where
     <C as Group>::ScalarField: Absorb,
 {
-    pub fn new(w: Vec<C::ScalarField>, e_len: usize) -> Self {
-        // note: at the current version, we don't use the blinding factors and we set them to 0
-        // always.
+    pub fn new(
+        w: Vec<C::ScalarField>,
+        rW: C::ScalarField,
+        e_len: usize,
+        rE: C::ScalarField,
+    ) -> Self {
         Self {
             E: vec![C::ScalarField::zero(); e_len],
-            rE: C::ScalarField::zero(),
+            rE,
             W: w,
-            rW: C::ScalarField::zero(),
+            rW,
         }
     }
+
     pub fn commit<CS: CommitmentScheme<C, H>, const H: bool>(
         &self,
         params: &CS::ProverParams,
@@ -282,8 +286,9 @@ where
 
 /// Implements Nova+CycleFold's IVC, described in [Nova](https://eprint.iacr.org/2021/370.pdf) and
 /// [CycleFold](https://eprint.iacr.org/2023/1192.pdf), following the FoldingScheme trait
+/// The `H` const generic specifies whether the homorphic commitment scheme is blinding
 #[derive(Clone, Debug)]
-pub struct Nova<C1, GC1, C2, GC2, FC, CS1, CS2, const H: bool>
+pub struct Nova<C1, GC1, C2, GC2, FC, CS1, CS2, const H: bool = false>
 where
     C1: CurveGroup,
     GC1: CurveVar<C1, CF2<C1>> + ToConstraintFieldGadget<CF2<C1>>,
@@ -323,7 +328,6 @@ where
     /// CycleFold running instance
     pub cf_W_i: Witness<C2>,
     pub cf_U_i: CommittedInstance<C2>,
-    cs_is_hiding: bool,
 }
 
 impl<C1, GC1, C2, GC2, FC, CS1, CS2, const H: bool> FoldingScheme<C1, C2, FC>
@@ -450,7 +454,6 @@ where
             // cyclefold running instance
             cf_W_i: cf_w_dummy.clone(),
             cf_U_i: cf_u_dummy.clone(),
-            cs_is_hiding: H,
         })
     }
 
@@ -462,6 +465,20 @@ where
         // Nova does not support multi-instances folding
         _other_instances: Option<Self::MultiCommittedInstanceWithWitness>,
     ) -> Result<(), Error> {
+        // ensure that commitments are blinding if user has specified so.
+        if H {
+            let blinding_commitments = if self.i == C1::ScalarField::zero() {
+                vec![self.w_i.rW, self.w_i.rE]
+            } else {
+                vec![self.w_i.rW, self.w_i.rE, self.W_i.rW, self.W_i.rE]
+            };
+            if blinding_commitments.contains(&C1::ScalarField::zero()) {
+                return Err(Error::IncorrectBlinding(
+                    H,
+                    format!("{:?}", blinding_commitments),
+                ));
+            }
+        }
         // `sponge` is for digest computation.
         let sponge = PoseidonSponge::<C1::ScalarField>::new(&self.poseidon_config);
         // `transcript` is for challenge generation.
@@ -677,7 +694,7 @@ where
         // set values for next iteration
         self.i += C1::ScalarField::one();
         self.z_i = z_i1;
-        self.w_i = Witness::<C1>::new(w_i1, self.r1cs.A.n_rows);
+        self.w_i = Witness::<C1>::new(w_i1, self.w_i.rW, self.r1cs.A.n_rows, self.w_i.rE);
         self.u_i = self.w_i.commit::<CS1, H>(&self.cs_pp, x_i1)?;
         self.W_i = W_i1;
         self.U_i = U_i1;
@@ -840,7 +857,7 @@ where
             self.cf_r1cs.clone(),
             self.cf_cs_pp.clone(),
             self.pp_hash,
-            cf_W_i,
+            cf_W_i.clone(),
             cf_U_i,
             cf_u_i_x,
             cf_circuit,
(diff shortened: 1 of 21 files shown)


folding-schemes/src/folding/nova/mod.rs
```diff
@@ -323,18 +323,19 @@ where
     /// CycleFold running instance
     pub cf_W_i: Witness<C2>,
     pub cf_U_i: CommittedInstance<C2>,
+    cs_is_hiding: bool,
```
arnaucube, line 326: Is this parameter actually needed? It seems that it could be removed, right? (also removing it from the serialization, since it is defined at the trait level)

folding-schemes/src/folding/nova/mod.rs
```diff
@@ -283,15 +283,15 @@ where
 /// Implements Nova+CycleFold's IVC, described in [Nova](https://eprint.iacr.org/2021/370.pdf) and
 /// [CycleFold](https://eprint.iacr.org/2023/1192.pdf), following the FoldingScheme trait
```
arnaucube, line 284: maybe adding a comment on what's `H`, as in the `CommitmentScheme` comment: [sonobe/folding-schemes/src/commitment/mod.rs#L13](https://github.com/privacy-scaling-explorations/sonobe/blob/3099befc6a7c7d9a3a35b083cacfb0fa393b2cd9/folding-schemes/src/commitment/mod.rs#L13)

folding-schemes/src/folding/nova/mod.rs
```diff
@@ -176,7 +176,7 @@ where
             rW: C::ScalarField::zero(),
```
arnaucube, line 176: Should the blinding factors of a new Nova `Witness` (`rE`, `rW`) be updated to support actual non-zero values?

(to something similar to what hypernova does for LCCCS & CCCS, eg. [sonobe/folding-schemes/src/folding/hypernova/lcccs.rs#L45](https://github.com/privacy-scaling-explorations/sonobe/blob/edadcdd520b8cf657cfa5c679dd09c6579759b0c/folding-schemes/src/folding/hypernova/lcccs.rs#L45) )

folding-schemes/src/folding/nova/mod.rs
```diff
@@ -283,15 +283,15 @@ where
 /// Implements Nova+CycleFold's IVC, described in [Nova](https://eprint.iacr.org/2021/370.pdf) and
 /// [CycleFold](https://eprint.iacr.org/2023/1192.pdf), following the FoldingScheme trait
 #[derive(Clone, Debug)]
-pub struct Nova<C1, GC1, C2, GC2, FC, CS1, CS2>
+pub struct Nova<C1, GC1, C2, GC2, FC, CS1, CS2, const H: bool>
```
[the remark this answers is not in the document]
dmpierre: Good idea, [done](https://github.com/privacy-scaling-explorations/sonobe/blob/19ca02989ad77d8b598374b74757cdc8a4dd7c52/folding-schemes/src/folding/nova/mod.rs#L291)!

folding-schemes/src/folding/nova/mod.rs
dmpierre: Good suggestion, added it [here](https://github.com/privacy-scaling-explorations/sonobe/blob/19ca02989ad77d8b598374b74757cdc8a4dd7c52/folding-schemes/src/folding/nova/mod.rs#L289)

folding-schemes/src/folding/nova/mod.rs
dmpierre: Yes you are right! In fact, I updated the PR quite a bit to make sure that those blinding values are set correctly when running `prove_step` for both [nova](https://github.com/privacy-scaling-explorations/sonobe/blob/2707b0cf8ec03a5ca8946eee59b179bb4f12b498/folding-schemes/src/folding/nova/mod.rs#L469) and [hypernova](https://github.com/privacy-scaling-explorations/sonobe/blob/2707b0cf8ec03a5ca8946eee59b179bb4f12b498/folding-schemes/src/folding/hypernova/mod.rs#L511)

CPerezz (approved):
LGTM modulo all arnaucube 's comments applied.

folding-schemes/src/folding/nova/mod.rs
```diff
@@ -831,13 +851,13 @@ where
...
     > {
-        fold_cyclefold_circuit::<C1, GC1, C2, GC2, FC, CS1, CS2>(
+        fold_cyclefold_circuit::<C1, GC1, C2, GC2, FC, CS1, CS2, H>(
             NOVA_CF_N_POINTS,
             transcript,
             self.cf_r1cs.clone(),
             self.cf_cs_pp.clone(),
             self.pp_hash,
-            cf_W_i,
+            cf_W_i.clone(),
```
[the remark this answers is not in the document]
dmpierre: Ah yes, good catch. Not sure why I did this in the first place in fact, thanks for spotting it! [6d8b487](https://github.com/privacy-scaling-explorations/sonobe/pull/129/commits/6d8b487a1ccd911966067ba297e5b2a0d96213a2)

folding-schemes/src/folding/nova/mod.rs
```diff
@@ -677,7 +694,7 @@ where
         // set values for next iteration
         self.i += C1::ScalarField::one();
         self.z_i = z_i1;
-        self.w_i = Witness::<C1>::new(w_i1, self.r1cs.A.n_rows);
+        self.w_i = Witness::<C1>::new(w_i1, self.w_i.rW, self.r1cs.A.n_rows, self.w_i.rE);
```
arnaucube, line 697: does this mean that `u_i.r_W` & `u_i.r_E` are propagated across all the folding steps (ie. $u_{i+1}.r_W==u_i.r_W$)? meaning for example that the `u_i.r_W` at step i=1 will be the same as in i=100?

arnaucube: Maybe it could be done as it is done in hypernova, not reusing the blinding values across all the instances, but using new randomness for each new instance: [https://github.com/privacy-scaling-explorations/sonobe/folding-schemes/src/folding/hypernova/mod.rs#L765](https://github.com/privacy-scaling-explorations/sonobe/blob/6248a90e89a6890af6603138586aa3e1fb683965/folding-schemes/src/folding/hypernova/mod.rs#L765)

folding-schemes/src/folding/nova/mod.rs
```diff
@@ -166,17 +166,21 @@ impl<C: CurveGroup> Witness<C>
...
+        e_len: usize,
+        rE: C::ScalarField,
+    ) -> Self {
         Self {
             E: vec![C::ScalarField::zero(); e_len],
-            rE: C::ScalarField::zero(),
+            rE,
             W: w,
-            rW: C::ScalarField::zero(),
+            rW,
```
arnaucube, line 179: if I understand correctly, this change has been introduced in the last update of the PR. Related to https://github.com/privacy-scaling-explorations/sonobe/pull/129/files#r1690865824 , could this be done as in hypernova? (and not reusing the blindings across different instances)

folding-schemes/src/folding/nova/mod.rs
dmpierre: That is the case, I was not too sure whether they should be re-used or not, but that should be better this way. [a06461f](https://github.com/privacy-scaling-explorations/sonobe/pull/129/commits/a06461f86ce97ca7250f60760c257f6e4822ce2a)

folding-schemes/src/folding/nova/mod.rs
dmpierre: Just commented about this above, tell me if that's what you had in mind!

arnaucube: This is a very minor thing, feel free to skip it:

as in https://github.com/privacy-scaling-explorations/sonobe/pull/129#discussion_r1682099466 and https://github.com/privacy-scaling-explorations/sonobe/pull/129#discussion_r1690865824 , I think this could be done as in HyperNova implementation for the blinding params, for example:
moving the current lines 694-702 into `Witness::new`
```rust
        // if hiding, sample new blinding values
        let (rW, rE) = if H {
            (
                C1::ScalarField::rand(&mut rng),
                C1::ScalarField::rand(&mut rng),
            )
        } else {
            (C1::ScalarField::zero(), C1::ScalarField::zero())
        };
```
so that in this way there is no need to pass the random values into the `Witness::new`, and is part of the 'new witness creation' logic to generate the blinding factors if hiding is activated.


The reasoning is having similar approach in both nova&hypernova, but also because there is no need to choose the random blindings outside the `Witness` creation, so that the `Witness::new` interface can be simplier and with less logic done before (the random values creation with the `if H`).

arnaucube (approved):
LGTM! Getting closer to the zk ^^

folding-schemes/src/folding/nova/mod.rs
dmpierre: Yes I agree, I did two commits that implement it this way. The first one adds a `const H: bool` generic to the `new` method when instantiating a new witness. The second one adds a new method for generating dummy witnesses. This is useful to avoid passing an `rng` object when instantiating dummy instances.

arnaucube (approved):

PR #141, opened 2024-08-09
add frontend features to reduce the amount of dependencies and compil…

### LINKED ISSUE #138: Add cargo features to disable by default the frontends
Add cargo features to disable by default the frontends, and then we can turn on the only specific frontend that we want to use (ie. Noir, Circom or Noname for the moment).

The reasoning behind this is that compilation time now is too long, in part because dependencies on specific frontends, so if the compilation don't use some frontend it can be turned off by default.

arnaucube:
…ation time

The reasoning behind this is that compilation time now is too long, in part because dependencies on specific frontends, so if the compilation don't use some frontend it can be turned off by default. Also in a normal usage of the lib the dev would choose one of the available frontends, but probably not all of them simultaneously.

resolves #138

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
         feature_set: [basic]
         include:
           - feature_set: basic
-            features: --features default,light-test
+            features: --features default,light-test,circom,noname,noir
     steps:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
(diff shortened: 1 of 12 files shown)


CPerezz (commented):
I expected Rust to not compile the deps that are not used within the final binary.

2 questions:
- Have you checked that by creating a new binary and adding sonobe as a dep only using one frontend, if all of them are compiled? I know that Rust has some smart optimizations that work towards this things.

- Do you have a sense of what improvements in compile time do we get with this PR? Would like to have the improvement noted somewhere.

PR #157, opened 2024-09-08
Traits for witnesses and committed instances

winderica:
Since #145 is a huge PR, I would like to separate it into several parts. This PR adds traits `Witness{Var}Ext` and `CommittedInstance{Var}Ext`, serving as abstractions for witnesses and committed instances in Nova, HyperNova and ProtoGalaxy.
Looking ahead, the methods in these traits are going to be used by the unified generic decider circuit in #145.

Also, the `hash` method in `CommittedInstance{Var}Ext` has a default implementation, so we can eliminate the duplicated code of `hash` for Nova's `CommittedInstance`, HyperNova's `LCCCS`, and ProtoGalaxy's `CommittedInstance`.
The signature of `hash` is also changed, where `z_0` and `z_i` are now slices rather than owned vectors.
In this way, we can avoid cloning `z_i` in augmented circuits, as `z_i` might be very large.

### DIFF
--- a/folding-schemes/src/folding/hypernova/circuits.rs
+++ b/folding-schemes/src/folding/hypernova/circuits.rs
@@ -1,6 +1,6 @@
 /// Implementation of [HyperNova](https://eprint.iacr.org/2023/573.pdf) circuits
 use ark_crypto_primitives::sponge::{
-    constraints::CryptographicSpongeVar,
+    constraints::{AbsorbGadget, CryptographicSpongeVar},
     poseidon::{constraints::PoseidonSpongeVar, PoseidonSponge},
     CryptographicSponge,
 };
@@ -14,6 +14,7 @@ use ark_r1cs_std::{
     fields::{fp::FpVar, FieldVar},
     groups::GroupOpsBounds,
     prelude::CurveVar,
+    uint8::UInt8,
     R1CSVar, ToConstraintFieldGadget,
 };
 use ark_relations::r1cs::{
@@ -41,6 +42,7 @@ use crate::folding::{
         CF1, CF2,
     },
     nova::get_r1cs_from_cs,
+    traits::CommittedInstanceVarExt,
 };
 use crate::frontend::FCircuit;
 use crate::utils::virtual_polynomial::VPAuxInfo;
@@ -52,19 +54,16 @@ use crate::{
 
 /// Committed CCS instance
 #[derive(Debug, Clone)]
-pub struct CCCSVar<C: CurveGroup>
-where
-    <C as CurveGroup>::BaseField: PrimeField,
-{
+pub struct CCCSVar<C: CurveGroup> {
     // Commitment to witness
     pub C: NonNativeAffineVar<C>,
     // Public io
     pub x: Vec<FpVar<CF1<C>>>,
 }
+
 impl<C> AllocVar<CCCS<C>, CF1<C>> for CCCSVar<C>
 where
     C: CurveGroup,
-    <C as ark_ec::CurveGroup>::BaseField: PrimeField,
 {
     fn new_variable<T: Borrow<CCCS<C>>>(
         cs: impl Into<Namespace<CF1<C>>>,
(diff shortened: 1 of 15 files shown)


winderica:
Just rebased this PR onto main, and it should be ready for review now!

folding-schemes/src/folding/traits.rs
```diff
@@ -0,0 +1,121 @@
...
+
+use crate::{transcript::Transcript, Error};
+
+use super::circuits::CF1;
+
+pub trait CommittedInstanceExt<C: CurveGroup> {
+    /// The in-circuit representation of the committed instance.
+    type Var: AllocVar<Self, CF1<C>> + CommittedInstanceVarExt<C>;
+    /// `hash` implements the committed instance hash compatible with the
+    /// in-circuit implementation from `CommittedInstanceVarOps::hash`.
```
arnaucube, line 19: is it `CommittedInstanceVarOps` or `CommittedInstanceVarExt`?

folding-schemes/src/folding/traits.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    /// Returns the commitments contained in the committed instance.
+    fn get_commitments(&self) -> Vec<C>;
+
+    /// Returns `true` if the committed instance is an incoming instance, and
+    /// `false` if it is a running instance.
+    fn is_incoming(&self) -> bool;
+
+    /// Checks if the committed instance is an incoming instance.
+    fn check_incoming(&self) -> Result<(), Error> {
```
arnaucube, line 52: Question, I haven't found any usage of `is_incoming` nor `check_incoming` at the current branch, are they methods needed for later at #145 ?

folding-schemes/src/lib.rs
```diff
@@ -43,6 +43,8 @@ pub enum Error {
     IVCVerificationFail,
     #[error("zkIVC verification failed")]
     zkIVCVerificationFail,
+    #[error("Committed instance is expected to an incoming (fresh) instance")]
```
arnaucube, line 46: maybe `[...] to *be* an incoming`?

arnaucube (approved):
Really cool! LGTM!

folding-schemes/src/folding/traits.rs
winderica: Nice question! I thought a bit about the naming of these traits and chose `Ext` because I saw it in some APIs provided by rust std, e.g., the [`FileExt`](https://doc.rust-lang.org/std/os/unix/fs/trait.FileExt.html) trait specifies methods that are going to be implemented in the [`File`](https://doc.rust-lang.org/std/fs/struct.File.html) struct.
I have also considered `Ops`, but I couldn't remember why I preferred `Ext` over it 😂

Now I realize that `Ext` may sound like it is for *extending* committed instances and witnesses, but these traits in fact have *core* methods that every struct should implement. So `Ops` should be a better choice!

folding-schemes/src/folding/traits.rs
winderica: In fact it will be used in the [`refactor-arith-trait` branch](https://github.com/winderica/sonobe/tree/refactor-arith-trait), which is planned to be the second step to #145, but I haven't opened a PR for that.

folding-schemes/src/folding/traits.rs
arnaucube: Thanks for the context. From my side I'm fine with `Ext`, makes sense.
Here in the comment I was referring to the in-code comment using `Ops` while the type everywhere else is `Ext`, so that the comment refers to an unexisting type.

folding-schemes/src/folding/traits.rs
arnaucube: Cool, sounds good!

folding-schemes/src/lib.rs
winderica: Fixed in https://github.com/privacy-scaling-explorations/sonobe/pull/157/commits/8047ccf79586630e09d3de69af75cb3be2ea9910 😄

folding-schemes/src/folding/traits.rs
winderica: Oops I didn't notice the `CommittedInstanceVarOps::hash` part and have already renamed `Ext` to `Ops`... Should I revert the commit?

arnaucube: Up to you, I'll be happy either way hehe. The original comment was just to point out the in-code comment not matching the used name. But happy with any of both names, so whatever one you prefer :)

dmpierre (approved):

CPerezz (approved):
LGTM!

folding-schemes/src/folding/hypernova/circuits.rs
```diff
@@ -127,41 +144,44 @@ where
...
-        let mut sponge = sponge.clone();
-        let U_vec = [
-            self.C.to_constraint_field()?,
-            vec![self.u],
-            self.x,
-            self.r_x,
-            self.v,
+impl<C: CurveGroup> AbsorbGadget<C::ScalarField> for LCCCSVar<C> {
+    fn to_sponge_bytes(&self) -> Result<Vec<UInt8<C::ScalarField>>, SynthesisError> {
+        unimplemented!()
```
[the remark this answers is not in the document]
winderica: I intentionally left `to_sponge_bytes` unimplemented for all instances in Nova, HyperNova, and ProtoGalaxy, because this method will not be used anywhere.
But I just found that this can be done with a single line of code, and you can check it in https://github.com/privacy-scaling-explorations/sonobe/pull/157/commits/2c9eb4c232944fedf974c88fd7d54b2b49557545 :)

PR #166, opened 2024-10-03
Resolve the stack overflow issue when evaluating polynomials in-circuit

### LINKED ISSUE #80: ark-r1cs-std interpolate_and_evaluate failing for n>11

winderica:
This PR uncomments the code for evaluating polynomials in-circuit, because the stack overflow issue was already addressed in https://github.com/arkworks-rs/r1cs-std/pull/145. There was another issue in `r1cs-std` that caused panics during polynomial interpolation, which is now fixed in https://github.com/winderica/r1cs-std/commit/70602229ba8c497e23619d3dd8a6be9b35db7c42 (corresponding PR: https://github.com/arkworks-rs/r1cs-std/pull/148).

Closes #80 and #130.

### DIFF
--- a/folding-schemes/src/folding/hypernova/decider_eth_circuit.rs
+++ b/folding-schemes/src/folding/hypernova/decider_eth_circuit.rs
@@ -26,7 +26,6 @@ use super::{
     nimfs::{NIMFSProof, NIMFS},
     HyperNova, Witness, CCCS, LCCCS,
 };
-use crate::{commitment::{pedersen::Params as PedersenParams, CommitmentScheme}, folding::nova::decider_eth_circuit::evaluate_gadget};
 use crate::folding::circuits::{
     cyclefold::{CycleFoldCommittedInstance, CycleFoldWitness},
     CF1, CF2,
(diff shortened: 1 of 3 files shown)


arnaucube (approved):
Awesome :raised_hands: !!!

Great work, and thanks for already porting the new fix to the cherry-picked r1cs-std branch so that we can already benefit from it!

CPerezz (approved):
LGTM!

PR #167, opened 2024-10-04
Add IVCProof to the existing folding schemes (Nova,HyperNova,ProtoGalaxy)

arnaucube:
Add IVCProof to the existing folding schemes (Nova,HyperNova,ProtoGalaxy), with serializers (by macro).

Implement `from_ivc_proof` for the FoldingSchemes trait (and Nova, HyperNova, ProtoGalaxy), so that the FoldingScheme IVC's instance can be constructed from the given parameters and the last IVCProof, which allows to sent the IVCProof between different parties, so that they can continue iterating the IVC from the received IVCProof.
Also the serializers allow for the IVCProof to be sent to a verifier that can deserialize it and verify it.

This allows to remove the logic from the file [folding/nova/serialize.rs](https://github.com/privacy-scaling-explorations/sonobe/blob/f1d82418ba047cf90805f2d0505370246df24d68/folding-schemes/src/folding/nova/serialize.rs) (removing the whole file), which is now covered by the `IVCProof` generated serializers (generated by macro instead of handwritten), and the test that the file contained is now abstracted and applied to all the 3 existing folding schemes (Nova, HyperNova, ProtoGalaxy) at the [folding/mod.rs](https://github.com/privacy-scaling-explorations/sonobe/blob/main/folding-schemes/src/folding/mod.rs) file.

It updates too the `IVC::verify` interface so that now it takes less parameters as inputs (basically the new IVCProof).

Also, update Nova VerifierParams serializers to avoid serializing the R1CS to save big part of the old serialized size (several MBs reduced).

The main improvement is that now we have the data structure of `IVCProof` which is portable (serializable) between different machines, so that one prover can send their current IVC state with the IVCProof to another one, and the second prover can continue the IVC folding, and eventually generate the Decider compressed zkSNARK proof for a verifier to verify it (or even send the plain IVCProof to a verifier). This also allows to remove the serializers of the Nova & HyperNova structs.
And also it adds a unified interface for deserializing the FoldingSchemes::ProverParam & VerifierParam.


folding-schemes/src/folding/protogalaxy/mod.rs
```diff
@@ -335,6 +399,55 @@ where
...
+    C2: CurveGroup,
+    CS1: CommitmentScheme<C1>,
+    CS2: CommitmentScheme<C2>,
+{
+    fn deserialize_with_mode<R: std::io::prelude::Read>(
+        mut _reader: R,
+        _compress: ark_serialize::Compress,
+        _validate: ark_serialize::Validate,
+    ) -> Result<Self, SerializationError> {
+        unimplemented!("use FoldingScheme::vp_deserialize_with_mode method instead");
```
dmpierre, line 447: What do you think of not implementing this trait at all?

folding-schemes/src/folding/nova/mod.rs
```diff
@@ -392,21 +377,11 @@ where
...
-            r1cs,
-            cf_r1cs,
-            cs_vp,
-            cf_cs_vp,
-        })
+        mut _reader: R,
+        _compress: ark_serialize::Compress,
+        _validate: ark_serialize::Validate,
+    ) -> Result<Self, SerializationError> {
+        unimplemented!("use FoldingScheme::vp_deserialize_with_mode method instead");
```
dmpierre, line 384: Should we not implement this trait at all eventually? This might help users avoid making this mistake.

folding-schemes/src/folding/protogalaxy/mod.rs
dmpierre, line 447: Same here, is it possible to not implement it at all?

dmpierre (approved):
This is neat! I have one question about keeping the implementation of a serializer trait. Apart from that, this PR removes quite a lot, nice!

folding-schemes/src/folding/nova/mod.rs
arnaucube: The thing is that previously it was serializing the whole R1CS of the main circuit and the R1CS of the CycleFold circuit, which was taking several tens/hundreds of MBs.
The new approach recomputes both R1CS on the fly, so that they don't need to be serialized, saving up tens/hundreds of MBs depending on the folding circuit. To do so, it needs the FCircuit but also it's initializing parameters, and also the GC1 and GC2, in order to rebuild it, but the GC1 & GC2 (CurveVar) don't allow for a serialization (as you also faced and mentioned in the PR https://github.com/privacy-scaling-explorations/sonobe/pull/107 ). So the way around has been to do the deserialization logic from the Nova struct itself, which can easily reconstruct the circuits from the given trait elements and thus recompute the two R1CS to rebuild the VerifierParams.

folding-schemes/src/folding/protogalaxy/mod.rs
arnaucube: https://github.com/privacy-scaling-explorations/sonobe/pull/167#discussion_r1796628236

folding-schemes/src/folding/nova/mod.rs
dmpierre: Makes sense, but it seems like this implementation can be removed without taking out any functionality, if not making users less confused?

arnaucube: Yes, but thought about giving a hit to the dev that uses the lib and expects to call the deserialize method, so that they know that it is not supported (hence the `unimplemented` message redirecting to the other method).

dmpierre: I understand, maybe showing devs how to serde could be documented instead of doing it with additional LOCs?

CPerezz (approved):
LGTM!! Nice work!!!

PR #176, opened 2024-11-19
migrate tests to use `?` instead of `unwrap` & `experimental-frontends` renaming

arnaucube:
- rename `frontends` to `experimental-frontends` to make more explicit that those are *experimental* and not recommended for usage, as discussed in the last meeting
  - the recommended frontend is defining the FCircuit with arkworks
- rename `upper_bound_ccs` to `compute_concrete_ccs` to describe better the method logic
  - thanks to @tremblaythibaultl for pointing out that the method name was not accurate
- migrate tests to use result instead of unwrap
  - motivation: in this way tests are like in-library code and there is no differenciation between test code and library code in terms of style, where errors are returned with `?` instead of `unwrap`. Same for examples. This was already done partially like this in some of the PrtoGalaxy tests, now on almost all the folding-schemes tests.


folding-schemes/src/lib.rs
```diff
@@ -122,6 +122,8 @@ pub enum Error {
     NoMultiInstances,
     #[error("Missing 'other' instances, since this is a multi-instances folding scheme. Expected number of instances, mu:{0}, nu:{1}")]
     MissingOtherInstances(usize, usize),
+    #[error("Type conversion error")]
+    TypeConversion,
```
winderica, line 126: Maybe it's better to combine this with https://github.com/privacy-scaling-explorations/sonobe/blob/8a41e678223028edf9c89d4a1901df2eabfed54c/folding-schemes/src/lib.rs#L117-L118

arnaucube: Oh right! Absolutely, updated: https://github.com/privacy-scaling-explorations/sonobe/pull/176/commits/22dd53df3b44bc9d1f725f9685e291c55de3183e

winderica: Awesome! Now everything looks good to me :)

winderica (approved):

PR #179, opened 2024-11-30
Bump arkworks to `0.5.0`

winderica:
Now most of the forks listed in #146 are eliminated, although two of them are still inevitable (for now):

```toml
ark-circom = { git = "https://github.com/winderica/circom-compat", branch = "arkworks-next", default-features = false }
acvm = { git = "https://github.com/winderica/noir", branch = "arkworks-next", default-features = false }
```


frontends/src/noir/bridge.rs
winderica, line 1: I am not sure if we should maintain this file and `noname/bridge.rs` in the sonobe repo or in separate repos (e.g. in @dmpierre's forks). I tentatively adopt the former approach but am also happy to change to the latter!

arnaucube (approved):
This is great! Not only migrating to the newer version but also cleaning the trait bounds and the code in general! Very nice!


Regarding the Noir frontend, I think that as you put it it's fine for now, and probably eventually a good move would be to move all the experimental-frontends in a separate repo (one of the reasons being, apart from being highly experimental and slow frontends at their current version, it has been one of the friction/delaying points when advancing the main sonobe crates, due newer versions on the dependencies of the dependencies of the frontends were released, or just by adding extra complexity). For this PR I think it's good as you done it.

Pratyush:
Sorry for reviving a dead PR. I see that y'all are using a fork of circom-compat; is there something we can do to help remove the need for the fork?

winderica:
Hey @Pratyush, thank you for reaching out (as well as your efforts in maintaining the arkworks ecosystem)!

At the time when I was working on this PR, I didn't realize that `circom-compat` had already been updated to 0.5.0, so I created a fork as a temporary workaround. Later, I found a 0.5.0 branch (which I guess has been merged into `main` now), but I had to focus on something else and didn't have time to work on the upgrade.

Today I decided to give it a try thanks to your message. Fortunately, the upgrade was overall smooth, and the only tricky part was to handle the breaking changes in `wasmer`. Now I just opened #221 that removes the dependency on my fork!

Pratyush:
That's fantastic to hear, and glad that you were able to upgrade smoothly!

PR #183, opened 2024-12-16
reduce HyperNova proving time, reduce computation in Nova&Mova compute_T method

arnaucube:
- add flamegraphs to hypernova&protogalaxy benchmarks to make it a bit easier to identify bottlenecks
- reduce ~65% proving time in HyperNova by removing the not-needed cs.finalize calls
- update compute_T (Nova & Mova) using Mova optimized approach from Mova's section 5.2 (https://eprint.iacr.org/2024/1220.pdf)

### DIFF
--- a/benches/hypernova.rs
+++ b/benches/hypernova.rs
@@ -1,4 +1,5 @@
 use criterion::*;
+use pprof::criterion::{Output, PProfProfiler};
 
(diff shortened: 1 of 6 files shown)


winderica (approved):
Awesome work. Looks great to me!

PR #184, opened 2024-12-17
Reduce time for native computation in `prove_step`

winderica:
This PR provides a tiny optimization of `prove_step` and avoids *natively* computing values that are also derived *in-circuit*. More specifically, sometimes we have two versions (native and in-circuit) of the same computation, and the native part is in fact unnecessary because we can extract the results from the circuit.

For example:
- Instead of running `step_native` to obtain the next state $z_{i + 1}$, we modify the augmented circuit which now returns the result of `generate_step_constraints`, and the native value $z_{i + 1}$ is now retrieved by calling `.value()` on the in-circuit variable.
  - A possible downside of removing `step_native` is that, if the user implements `generate_step_constraints` and `step_native` with inconsistent behaviors (e.g., when `step_native` is correct but `generate_step_constraints` has bugs), then $z_{i + 1}$ will become incorrect. For this, we may recommend the user to write unit tests for their step circuits.
- Instead of computing public inputs natively for the augmented circuit and the CycleFold circuit, we do the computation solely in circuit and mark the results as public.

The performance gain is not very significant because `step_native` is not the bottleneck, but the additional benefit is that we can have cleaner and less duplicated code.

### DIFF
--- a/folding-schemes/src/folding/circuits/cyclefold.rs
+++ b/folding-schemes/src/folding/circuits/cyclefold.rs
@@ -9,6 +9,7 @@ use ark_r1cs_std::{
     eq::EqGadget,
     fields::fp::FpVar,
     prelude::CurveVar,
+    R1CSVar,
 };
 use ark_relations::r1cs::{
     ConstraintSynthesizer, ConstraintSystem, ConstraintSystemRef, Namespace, SynthesisError,
@@ -443,8 +444,6 @@ pub struct CycleFoldCircuit<CFG: CycleFoldConfig, GC: CurveVar<CFG::C, CF2<CFG::
     pub r_bits: Option<Vec<bool>>,
     /// points to be folded in the CycleFoldCircuit
     pub points: Option<Vec<CFG::C>>,
-    /// public inputs (cf_u_{i+1}.x)
-    pub x: Option<Vec<CF2<CFG::C>>>,
 }
 
 impl<CFG: CycleFoldConfig, GC: CurveVar<CFG::C, CF2<CFG::C>>> CycleFoldCircuit<CFG, GC> {
@@ -454,7 +453,6 @@ impl<CFG: CycleFoldConfig, GC: CurveVar<CFG::C, CF2<CFG::C>>> CycleFoldCircuit<C
             _gc: PhantomData,
             r_bits: None,
             points: None,
-            x: None,
         }
     }
 }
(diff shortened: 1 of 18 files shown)


folding-schemes/src/folding/circuits/cyclefold.rs
```diff
@@ -520,13 +512,19 @@ impl<CFG: CycleFoldConfig, GC: CurveVar<CFG::C, CF2<CFG::C>>> ConstraintSynthesi
...
         ]
         .concat();
-        computed_x.enforce_equal(&x)?;
+        #[cfg(test)]
+        assert_eq!(x.len(), CFG::IO_LEN); // non-constrained sanity check
+
+        // This line "converts" `x` from a witness to a public input.
+        // Instead of directly modifying the constraint system, we explicitly
+        // allocate a public input and enforce that its value is indeed `x`.
+        Vec::new_input(cs.clone(), || x.value())?.enforce_equal(&x)?;
```
winderica, line 527: This (and other similar lines) may look weird because `x` is compared with itself. In fact, `x.value()` here allows an honest prover to extract the public inputs, and `enforce_equal` prevents a malicious prover from claiming incorrect public inputs that are different from the honest `x` computed by the circuit.

arnaucube: Thanks for the comment! maybe this comment could be added to the inline code comment? (the reasoning about the honest prover & malicious prover).
(Same for the other occurrences of this line in `{hypernova, nova, protogalaxy}/circuits.rs`)

arnaucube (commented):
This is a very good idea! Very nice ^^
And not only saves the step_native call but also several hashes along the prove_step, awesome!

Now that it is no longer needed, to avoid missusage the [`FCircuit.step_native`](https://github.com/winderica/sonobe/blob/089365c53d4818ecacdf035740947023433d8a3e/folding-schemes/src/frontend/mod.rs#L29) method should be removed right?

winderica:
> Now that it is no longer needed, to avoid missusage the [`FCircuit.step_native`](https://github.com/winderica/sonobe/blob/089365c53d4818ecacdf035740947023433d8a3e/folding-schemes/src/frontend/mod.rs#L29) method should be removed right?

Good point! Will remove this method from the `FCircuit` trait.

winderica:
> > Now that it is no longer needed, to avoid missusage the [`FCircuit.step_native`](https://github.com/winderica/sonobe/blob/089365c53d4818ecacdf035740947023433d8a3e/folding-schemes/src/frontend/mod.rs#L29) method should be removed right?
> 
> Good point! Will remove this method from the `FCircuit` trait.

Done! While `FCircuit::step_native` is removed now, there is still some code that relies on native step functions for test purposes, which I think is fine.

arnaucube (approved):
Really really great! Very nice improvement!

PR #188, opened 2024-12-20
chore: docs + lint

Autoparallel:
## In this PR
Here's what I'm doing:

### Docs
Working to adding more documentation adhering to the crate level attributes:
```
#![warn(missing_docs, clippy::missing_docs_in_private_items)]
```
Personally, I enjoy having these lint warnings appear as they keep me honest in keeping well documented code. I do have to say, Sonobe is pretty well documented already! But this has helped me find some spots to add to, and `docs_in_private_items` is really nice for other developers.

### Lint
I usually have my clippy set to `pedantic` so I get a bunch of... pedantic lint warnings! These have helped me catch things like redundant clones, methods that don't need to return `Result`, and also catch spots where we cast into `usize` from types like `u64`.

---

This PR is still a WIP. I will continue to add to it and flush out the `folding-schemes` crate prior to marking ready to review.


Autoparallel:
Leaving off for here now.

Autoparallel:
Pinging here real quick. 

Note commits `4779599` and `56f3063` I make some internal API changes that remove a good amount of cloning. I suspect  that this could be taken further, but it might be smarter to cherry pick these into a different branch and focus this PR on docs.

What do maintainers think?

folding-schemes/src/commitment/kzg.rs
```diff
@@ -93,20 +129,12 @@ where
...
-    /// be computed in the most trustless way possible, usually through a MPC ceremony.
     fn setup(
         mut rng: impl RngCore,
         len: usize,
     ) -> Result<(Self::ProverParams, Self::VerifierParams), Error> {
         let len = len.next_power_of_two();
+        // TODO (autoparallel): This `expect` will panic when the function itself is capable of returning an error.
         let universal_params =
             KZG10::<E, DensePolynomial<E::ScalarField>>::setup(len, false, &mut rng)
                 .expect("Setup failed");
```
arnaucube, line 140: Right, agree, looks like by replacing the expect by `?` should work (?).

folding-schemes/src/commitment/kzg.rs
```diff
@@ -70,19 +81,44 @@ impl<'a, C: CurveGroup> Valid for ProverKey<'a, C> {
...
+    /// Inner [`Pairing`] friendly curve
     _e: PhantomData<E>,
 }
 
+/*
+TODO (autoparallel): Okay, I'm noticing something here, basically I think that there should likely be two implementations for `CommitmentScheme<G>`,
+one that is hiding, and one that is not (as opposed to the const generic `H` in the trait itself). We could have `HidingCommitmentScheme: CommitmentScheme`.
+If impl `CommitmentScheme<G> for KZG<'a, E, H>`, then we can impl `HidingCommitmentScheme for KZG<'a, E, true>`. The implementation of `HidingCommitmentScheme`
+would be super straight forward as it would just add the blinding factor to the output of the "super" `CommitmentScheme` `commit` and `prove` methods. Then those
+methods on `CommitmentScheme` do not have to take in `blind: Option<E::ScalarField>` or the `dyn Rng`.
```
arnaucube, line 119: Would this mean having almost the same code twice for each the commitment scheme twice? (one struct for hiding and one struct for non hiding)
With the `H` param we reuse the (almost) same impl while both (hiding &non-hiding) are available.

From the using point of view, since the `H` is set to default `H=false`, if when using the struct it is not specified (just not setting any `H` value), it acts like is `false`.
In other words,
- the `NonHidingCommitmentScheme` currently is achieved by `CommitmentScheme<G1>` (where `H=false` by default)
- and the `HidingCommitmentScheme` is achieved by `CommitmentScheme<G1, true>`

folding-schemes/src/folding/circuits/mod.rs
```diff
@@ -8,6 +41,9 @@ pub mod nonnative;
 pub mod sum_check;
 pub mod utils;
 
+// TODO (autoparallel): nit -- I think the short variable names `CF1` and `CF2` could be made more clear by something
+// like `ConstraintFieldPrimary` and `ConstraintFieldSecondary`, but this isn't a hill I'll die on
```
arnaucube, line 45: Agree on that their names are not super clear at first sight since it's an acronym name, the fear is that those are used a lot along the code and having a short name helps with code readability :S
(used a lot and in combined cases, like `CurveVar<C1, CF2<C1>>`)
So far we've tried to have an explanation of each of them in the comment of their definition, so that it leaves not doubt about what are each of them.

folding-schemes/src/folding/mod.rs
```diff
@@ -1,3 +1,38 @@
...
+//! - Leverages common circuit gadgets from the [`circuits`] module
+//!
+//! # References
+//!
+//! The implementations are based on the following academic works:
+//!
+//! - [Nova: Recursive Zero-Knowledge Arguments from Folding Schemes](https://eprint.iacr.org/2021/370)
+//! - [HyperNova: Recursive Arguments for Customizable Constraint Systems](https://eprint.iacr.org/2023/573)
+//! - [CycleFold: Folding-scheme-based Recursive Arguments over Different Curves](https://eprint.iacr.org/2023/1192)
+
```
arnaucube, line 35: also ProtoGalaxy ^^
`[ProtoGalaxy: Efficient ProtoStar-style folding of multiple instances](https://eprint.iacr.org/2023/1106)`

folding-schemes/src/folding/mod.rs
```diff
@@ -51,36 +80,41 @@ pub mod tests {
...
+        let f_circuit = FC::new(())?;
+
+        let prep_param = NovaPreprocessorParam::new(poseidon_config.clone(), f_circuit);
+        test_serialize_ivc_opt::<G1, G2, FC, N>("nova", &prep_param)?;
+        test_serialize_ivc_opt::<G1, G2, FC, HN>("hypernova", &prep_param)?;
+
         let prep_param = (poseidon_config, f_circuit);
-        test_serialize_ivc_opt::<G1, G2, FC, P>("protogalaxy".to_string(), prep_param)?;
+        test_serialize_ivc_opt::<G1, G2, FC, P>("protogalaxy", &prep_param)?;
```
arnaucube, line 95: Oh nice, this reorg makes it more clear!

folding-schemes/src/lib.rs
```diff
@@ -1,6 +1,51 @@
...
+//! 5. Perform folding operations via `prove_step`
+//!
+//! # References
+//!
+//! The library implements protocols from several academic works:
+//!
+//! * [Nova: Recursive Zero-Knowledge Arguments from Folding Schemes](https://eprint.iacr.org/2021/370)
+//! * [HyperNova: Recursive Arguments for Customizable Constraint Systems](https://eprint.iacr.org/2023/573)
+//! * [CycleFold: Folding-scheme-based Recursive Arguments over Different Curves](https://eprint.iacr.org/2023/1192)
+
```
arnaucube, line 44: also ProtoGalaxy ^^
`[ProtoGalaxy: Efficient ProtoStar-style folding of multiple instances](https://eprint.iacr.org/2023/1106)`

folding-schemes/src/lib.rs
```diff
@@ -1,6 +1,51 @@
+//! Sonobe is a library implementing various folding schemes for recursive SNARKs (Succinct Non-interactive ARguments of Knowledge)
+//! and IVC (Incremental Verifiable Computation). It provides a modular, extensible framework for working with different folding
+//! schemes including Nova, HyperNova, and other variants.
```
arnaucube, line 3: Probably worth mentioning ProtoGalaxy here.
The main 3 folding schemes we have now are Nova,HyperNova,ProtoGalaxy, the 3 of them with CycleFold, and additionally there are a couple of variations of Nova which are not finished yet but probably no need to mention them already here, which are Mova and Ova (variations in the sense that they are Nova with a modification in how to deal with the E term of the RelaxedR1CS).

folding-schemes/src/commitment/pedersen.rs
```diff
@@ -77,6 +92,7 @@ impl<C: CurveGroup, const H: bool> CommitmentScheme<C, H> for Pedersen<C, H> {
         Ok(params.h.mul(r) + C::msm_unchecked(&params.generators[..v.len()], v))
     }
 
+    // TODO (autoparallel): I'm guessing `_rng` is marked with prefix `_` because it goes unused always and it is planned for the future? Otherwise we should remove that prefix.
```
arnaucube, line 95: In this case (Pedersen), it is not used, but in an other implementation of the `CommitmentScheme` trait (ie. IPA) it is used.

arnaucube (commented):
Thanks for this PR, it really improves the code quality and it's docs! Also the removal of several clones is a nice improvement.

Left some comments, some of them are more questions to understand better the possibilities.

Autoparallel:
@arnaucube thank you for the feedback! Let me address some of the comments about docs more generally prior to addressing the review comments directly.

## Overview
In the `lib.rs` we have set some crate-level attributes:
```rust
#![allow(non_snake_case)]
#![allow(non_upper_case_globals)]
#![allow(non_camel_case_types)]
#![warn(missing_docs, clippy::missing_docs_in_private_items)] // new from me
```
With clippy running in CI and failing with warnings (or, we can set up `cargo-deny`, but I digress), if docs are missing from public items, we will trigger a warning via:
```rust
#![warn(missing_docs)]
```
and the second `#![warn(clippy::missing_docs_in_private_items)]` is a more aggressive lint on my part that throws warning when private items miss docs. The latter is really only useful for developers of the crate itself, right? The former is useful for anyone using the crate. 

## Rust Style
The general Rust style so that docs.rs is consistent usually relies on what `missing_docs` throws. These docs are the same as what you see in hover-over in an IDE. Anything marked `pub`, including fields of a struct or variants of a `pub enum` then require docs. Of course, you don't have to do this, but I do think it adds a nice touch. You are right though that sometimes these docs are then a bit unnecessary, but (and this is my personal preference -- you're entitled to yours!) I'd rather have some unnecessary docs but a fully documented crate than one that lacks docs.

Furthermore, cargo docs do have some implicit formatting structure guidance. These include:
```rust
#![warn(clippy::missing_errors_doc)]
#![warn(clippy::missing_panic_doc)]
#![warn(clippy::missing_safety_doc)]
```
Respectively, these trigger if you don't include why a function that returns `Result` may error, why a function that can panic would panic, and why a use of `unsafe` is valid. 

You can go as to others such as as:
```rust
#![warn(rustdoc::missing_doc_code_examples)]
```
to trigger when a public component doesn't also provide an example of usage.

## Final Comments
My suggestion is to have a lint workflow (this repo already does) and to set warning attributes for what is deemed acceptable. I do think the two I added are nice and I am happy to finish this crate with it completely. Moving forward, it would then force contributors to write docs for what they add or change. The extra cargo/rustdoc style lints are maybe too pedantic especially to drop all at once, but are also worth consideration in my mind.

If at the end of the day we prefer not to lint these at all, I still use them when I'm writing the docs to tell me where I need them :)

---
As a side note, in my rust analyzer, I have set clippy to be [pedantic](https://doc.rust-lang.org/clippy/lints.html#pedantic). I think this is overkill for a crate itself, but it does point towards some things. It may be interesting for others to try this setting to see what kind of stuff gets caught in their crates!

Autoparallel:
Ah nice! The other cleanup PR got merged. I've been sick and have not had time to work on this, but I suppose that means I'll have an easier rebase. Will continue on with this after rebasing the changes from winderica

PR #158, opened 2024-09-11
Improve the design of circuit, config, and instances for CycleFold

winderica:
Depends on https://github.com/privacy-scaling-explorations/sonobe/pull/123. Before https://github.com/privacy-scaling-explorations/sonobe/pull/123 gets merged, a diff can be found here: https://github.com/privacy-scaling-explorations/sonobe/compare/winderica:protogalaxy-based-ivc...winderica:refactor-cyclefold.

This PR aims to refactor the existing code related to CycleFold improve the design. Specifically, it:
- [ ] Adds support for multiple random values when accumulating points in CF circuit, while maintaining the efficiency of current implementation for multiple points. (See the comments in https://github.com/privacy-scaling-explorations/sonobe/pull/123#discussion_r1748939022)
- [x] Extract the construction of `cf_u.x` into a separate function. Before this PR, `cf_u.x` is manually constructed by concatenating randomness and points in all augmented circuits of Nova, HyperNova, and ProtoGalaxy, as observed in https://github.com/privacy-scaling-explorations/sonobe/pull/123#discussion_r1736247883.
- [ ] ...

PR #189, opened 2024-12-21
Refactor and simplify the design of traits

### LINKED ISSUE #170: `to_native_sponge_field_elements` cant be called with PrimeField
In the `to_sponge_field_elements<F: PrimeField>(&self, dest: &mut Vec<F>) ` function we cannot call `to_native_sponge_field_elements(dest)` directly, as `to_native_sponge_field_elements` needs `F` to be `C::ScalarField`, but here `F` is a generic `PrimeField`. This affects all the folding schemes that use a `CryptographicSponge` (Nova, OVA, Hypernova, and the upcoming Mova)

winderica:
The current codebase has pretty complicated trait bounds / type parameters, and this is mostly due to the requirement of types that are more specific / powerful than the ones provided by arkworks.

For example, when using `CurveGroup`, we often need its `BaseField` to be a `PrimeField`, its `ScalarField` and `BaseField` to be `Absorb`able, and an in-circuit type `CurveVar` to be associated with the curve.

This PR aims to simplify the interfaces of complex traits, thereby improving the DX of sonobe before the incoming release. Specifically:
- Two new traits, `SonobeField` and `SonobeCurve`, are introduced, in order to remove type parameters like `GC{1, 2}: CurveVar` and trait bounds like `Absorb` and `PrimeField`.
- Traits and implementations related to sponges / transcripts are redesigned.
  - Type parameters for `AbsorbNonNative` are adjusted in order to fix #170.
  - Instead of implementing and calling `to_constraint_field` to obtain the hash preimage that corresponds to an in-circuit variable, `to_{native_}sponge_field_elements` is now preferred because it is semantically more meaningful.
    (Note that we cannot eliminate `to_constraint_field` calls for `CurveGroup` because arkworks does not implement `to_sponge_field_elements` for it, and it is also impossible for us to do the implementation as both traits are foreign.)
- The `Inputize` trait was originally added to replace arkworks' `ToConstraintField::to_field_elements`, since we have too many methods similar to the latter, whose naming is also less clearer than the former.
  Now, this PR separates the "inputization" of non-native variables from native variables, resulting in a new trait `InputizeNonNative`. The benefit is that. we can now simultaneously handle types with two representations in-circuit (e.g., `CurveGroup` whose in-circuit version can be either `CurveVar` or `NonNativeAffineVar`).
  Furthermore, `Inputize` for a type `T` is now *guaranteed* to return underlying field elements in the same order as how `T::Var` is allocated in-circuit. In comparison, arkworks' implementation of `ToConstraintField::to_field_elements` for EC group elements (which essentially returns `x`, `y`, `infinity` of the affine form) in fact does not match the allocation of in-circuit EC variables (which is handled in the order of `x`, `y`, `z` of the projective form).
- To allow easy conversion to EVM calldata, this PR implements the `ToEth` trait for arkworks types, so that we can avoid verbosity when implementing `prepare_calldata` for on-chain deciders for HyperNova, ProtoGalaxy, etc. in the future.
- The CycleFold-related code is also refactored. I think this is more related to #158 which I am still working on (sorry for the long delay!), but let me keep it as is in this PR.

Thanks in advance for reviewing this (yet another) huge PR. It would be great to hear your feedback!

### DIFF
--- a/folding-schemes/src/lib.rs
+++ b/folding-schemes/src/lib.rs
@@ -145,8 +145,8 @@ pub enum Error {
 ///
 /// In other words, C1.Fq == C2.Fr, and C1.Fr == C2.Fq.
 pub trait FoldingScheme<
-    C1: SonobeCurve<BaseField = C2::ScalarField, ScalarField = C2::BaseField>,
-    C2: SonobeCurve,
+    C1: Curve<BaseField = C2::ScalarField, ScalarField = C2::BaseField>,
+    C2: Curve,
     FC: FCircuit<C1::ScalarField>,
 >: Clone + Debug
 {
@@ -223,8 +223,8 @@ pub trait FoldingScheme<
 /// Trait with auxiliary methods for multi-folding schemes (ie. HyperNova, ProtoGalaxy, etc),
 /// allowing to create new instances for the multifold.
 pub trait MultiFolding<
-    C1: SonobeCurve<BaseField = C2::ScalarField, ScalarField = C2::BaseField>,
-    C2: SonobeCurve,
+    C1: Curve<BaseField = C2::ScalarField, ScalarField = C2::BaseField>,
+    C2: Curve,
     FC: FCircuit<C1::ScalarField>,
 >: Clone + Debug
 {
@@ -250,8 +250,8 @@ pub trait MultiFolding<
 }
 
 pub trait Decider<
-    C1: SonobeCurve<BaseField = C2::ScalarField, ScalarField = C2::BaseField>,
-    C2: SonobeCurve,
+    C1: Curve<BaseField = C2::ScalarField, ScalarField = C2::BaseField>,
+    C2: Curve,
     FC: FCircuit<C1::ScalarField>,
     FS: FoldingScheme<C1, C2, FC>,
 >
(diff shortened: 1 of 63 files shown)


folding-schemes/src/lib.rs
```diff
@@ -295,3 +307,28 @@ where
...
+    PrimeField<BasePrimeField = Self> + Absorb + AbsorbNonNative + Inputize<Self>
+{
+    type Var: FieldVar<Self, Self>;
+}
+
+impl<P: FpConfig<N>, const N: usize> SonobeField for Fp<P, N> {
+    type Var = FpVar<Self>;
+}
+
+pub trait SonobeCurve:
```
arnaucube, line 321: How would you see naming them `Field` and `Curve` directly?
The reasoning I'm thinking is that it's shorter than `SonobeField` and `SonobeCurve`, which since it's a name that is used a lot across the code might be a bit tedious having always the prefix `Sonobe` there. And also that the 'Sonobe' prefix can be assumed since those traits are defined inside the Sonobe crate context.

arnaucube (commented):
Very nice! Simplifies a lot the code, and the Absorb related logic and Inputize is more coherent now! Very cool!

folding-schemes/src/lib.rs
winderica: Makes sense! Will rename both.

arnaucube (approved):
Again, very cool improvements!

PR #191, opened 2024-12-25
Update the FCircuit trait, so that it supports custom data structures for the external inputs.

arnaucube:
Update the FCircuit trait, so that it supports custom data structures for the external inputs.

This also eliminates the need of having the `external_inputs_len` method in the `FCircuit`.

The motivation for this change is that in most practical use cases, the external inputs have a not-naive structure, and the old interface required that to convert the external inputs structure into a vector of finite field elements, which inside the FCircuit would be converted back into a custom data structure. This is specially tedious when dealing with curve points, which converting them from point to field elements and back to the point (both outside (rust native) and inside the circuit (constraints)) is a bit cumbersome.  With this update, it's much more straight forward to define the FCircuit with custom external inputs data structures.

For the experimental-frontends, the external inputs keep being an array of field elements.

> When reviewing this PR: the main change is in the file [folding-schemes/src/frontend/mod.rs](https://github.com/privacy-scaling-explorations/sonobe/pull/191/files#diff-b36294c106971af264dff9fa104e0e4602d6d3a540bc44cc9b7f130e4c3b06fe), most of the rest of the git diff is just updating the code to the new `FCircuit` interface.

### Context/motivation
When doing more 'real world' use cases, it is convenient to don't have the limitation to just arrays of field elements for the external inputs on the FCircuit. For example, when dealing with elliptic curve points, it required to convert the points to affine to get their coordinates as field elements, then concatenate the various field elements (coordinates) from various points to pass it as `Vec<FpVar<F>>` as the `external_inputs` to the FCircuit, which then inside the FCircuit that vector gets decomposed back into separated coordinates to rebuild the elliptic curve points (which in order to do so, need to break some of the abstractions to be able to build the specific curve points).
This grows in complexity as the data structures needed are more rich.

As an example, suppose we want to have as external inputs the following structure `ExtInpVar`:
```rust
pub struct ExtInpVar<C: CurveGroup, GC: CurveVar<C, CF<C>>> {
    msg: FpVar<CF<C>>,
    pk: GC,
    sig_r: GC,
    sig_s: Vec<Boolean<CF<C>>>,
}
```

with this PR, in the `FCircuit` we just need to set the trait parameters:
```rust
type ExternalInputs = ExtInp<C>;
type ExternalInputsVar = ExtInpVar<C, GC>;
```
And then we can just directly use that type for the `external_inputs`:
```rust
    fn generate_step_constraints(
        &self,
        cs: ConstraintSystemRef<F>,
        _i: usize,
        z_i: Vec<FpVar<F>>,
        external_inputs: Self::ExternalInputsVar,
    ) -> Result<Vec<FpVar<F>>, SynthesisError> {
        let mut count = z_i[0].clone();
        let res = verify::<C, GC>(
            cs.clone(),
            self.config.clone(),
            external_inputs.pk,
            (external_inputs.sig_r, external_inputs.sig_s),
            external_inputs.msg,
        )?;
        res.enforce_equal(&Boolean::<F>::TRUE)?;
        count = count.clone() + FpVar::<F>::one();

        Ok(vec![count])
    }
}
```

The same use case before this PR, would have required much more code, and furthermore it would have needed to remove some of the curve abstractions in order to be able to reconstruct the `GC` points from their coordinates inside the `FCircuit`.

Full use case example code available at: [https://github.com/arnaucube/fold-babyjubjubs/blob/main/src/fcircuit.rs#L53](https://github.com/arnaucube/fold-babyjubjubs/blob/4dfb2781421f3e98fc0556a13f0f4bac5d9c3eaa/src/fcircuit.rs#L53)

### DIFF
--- a/examples/external_inputs.rs
+++ b/examples/external_inputs.rs
@@ -74,6 +74,8 @@ where
     F: Absorb,
 {
     type Params = PoseidonConfig<F>;
+    type ExternalInputs = VecF<F>;
+    type ExternalInputsVar = VecFpVar<F>;
 
     fn new(params: Self::Params) -> Result<Self, Error> {
         Ok(Self {
@@ -84,9 +86,6 @@ where
     fn state_len(&self) -> usize {
         1
     }
-    fn external_inputs_len(&self) -> usize {
-        1
-    }
     /// generates the constraints and returns the next state value for the step of F for the given
     /// z_i and external_inputs
     fn generate_step_constraints(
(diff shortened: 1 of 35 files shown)


examples/external_inputs.rs
```diff
@@ -74,6 +74,8 @@ where
     F: Absorb,
 {
     type Params = PoseidonConfig<F>;
+    type ExternalInputs = VecF<F>;
+    type ExternalInputsVar = VecFpVar<F>;
```
winderica, line 78: I might be too nitpicky, but do you think it is a good idea to replace `VecF<F>` and `VecFpVar<F>` with `[F; 1]` and `[FpVar<F>; 1]`, so we can remove `VecF` and `VecFpVar` defined in this file (not in `experimental-frontends/src/utils.rs`) and make the example simpler?

Although `[FpVar<F>; 1]` does not satisfy `Default`, we can in fact remove the `Default` trait bound from
https://github.com/privacy-scaling-explorations/sonobe/blob/0dde06067885b394f7cb9b291df6c5a5d72fd130/folding-schemes/src/frontend/mod.rs#L20
because we can call `ExternalInputsVar::new_witness(cs.clone(), || Ok(external_inputs))?` instead of `ExternalInputsVar::default()` when we need dummy external inputs.

winderica (approved):
Awesome work! I like the design of `ExternalInputs` and `ExternalInputsVar`, which makes the step circuits more flexible!

> For the experimental-frontends, the external inputs keep being an array of field elements.

It is sad that rust does not implement `Default` for `[T; N]` where `T` already satisfies `Default`, and we need to wrap arrays of field elements by `VecF` and `VecFpVar`. But anyway the changes still look good to me!

folding-schemes/src/frontend/mod.rs
```diff
@@ -10,8 +10,14 @@ pub mod utils;
 /// inside the agmented F' function).
 /// The parameter z_i denotes the current state, and z_{i+1} denotes the next state after applying
 /// the step.
+/// Note that the external inputs for the specific circuit are defined at the implementation of
+/// both `FCircuit::ExternalInputs` and `FCircuit::ExternalInputsVar`, where the `Default` trait
+/// implementation. For example if the external inputs are just an array of field elements, their
```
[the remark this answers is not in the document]
arnaucube: omg, right, my bad, thanks for pointing it out! forgot to finish the comment sentence, updated at https://github.com/privacy-scaling-explorations/sonobe/pull/191/commits/7682fe0fa190fa3a229eb5e5d5e7d8e35e034734

examples/external_inputs.rs
arnaucube: Absolutely! This makes a lot of sense and it's more clean, updated it: https://github.com/privacy-scaling-explorations/sonobe/pull/191/commits/7682fe0fa190fa3a229eb5e5d5e7d8e35e034734
Thank you very much!

PR #193, opened 2025-01-10
Remove `fs` from the parameters of `Decider::preprocess`

winderica:
In the current `Decider` trait, the `preprocess` method takes `fs`, an instance of the `FoldingScheme` trait, as input.

However, note that `fs` is created by the prover by running `FoldingScheme::init` with the initial state `z_0`. On the other hand, `Decider::preprocess` is usually invoked by a trusted party at the beginning (before provers and verifiers are involved), when `z_0` is still unknown.

Thus, it makes more sense to remove the `fs` parameter from `Decider::preprocess`. The implementations of `Decider` are updated correspondingly, so that the preprocessing can be done without a concrete `fs`. Instead, we simply build a dummy decider circuit, during which we only need to provide some parameters that are known before the creation of `fs`.


arnaucube (approved):
Nice! More clean usage of the Decider now!

PR #195, opened 2025-01-14
Introduce getter methods for information related to constraint systems

winderica:
This PR introduces getter methods that return information about R1CS and CCS. With these getters, we can avoid declaring field members such as `ccs.m` as public.

Note that the `Arith` trait is now solely for these methods which are about the constraint system *itself*. The original `Arith` trait has been renamed to `ArithRelation`, stressing that it is about the *relation* between the constraint system's witness `W` and instance `U`.


arnaucube (approved):
LGTM!

PR #200, opened 2025-01-28
Update LICENSE

maximevtush:
Updated the copyright year from 2023 to 2025 in the LICENSE file


arnaucube:
hi @maximevtush , so far we don't see the need to keep updating the license file every year to update a number; seems fine with the year number that appears being the year of creation of the project. As much maybe we could setup a github CI that auto updates it in some way, but for sure not depending on manually doing it every year.

PR #202, opened 2025-02-03
Fixed bugs with the ptvsline implementation 

NiDimi:
1. As part of protocol 4 in the Mova [paper](https://eprint.iacr.org/2024/1220.pdf) the verifier and the prover need to agree on the rE_prime value (Step 3). This PR enforces these checks and adds tests to make sure that it cannot succeed without having the same one.
2. l(x) was calculated l(x) = r1 + x(**r1-r2**) instead of l(x) = r1 + x(**r2-r1**). The reason the protocol was working was because we were testing with the error term was always zero which resulted in the check h2(1) = ci2.mleE being successful when it shouldn't. Also added tests to catch issues like that in the future.


arnaucube (approved):
Cool, thanks!

PR #208, opened 2025-02-18
Fix git dependencies

pmikolajczyk41:
We should rely on fixed revs, instead of branches to avoid silent breakage / uncontrolled changes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ resolver = "2"
 
 
 [patch.crates-io]
-ark-r1cs-std = {git = "https://github.com/yelhousni/r1cs-std", branch="perf/sw"}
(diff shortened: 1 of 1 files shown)


winderica (approved):
Agreed! Thank you for your comment in https://github.com/privacy-scaling-explorations/sonobe/issues/146#issuecomment-2301839849 as well as this PR that addresses it!

In fact, we didn't switch to fixed revs because we hope to replace all forks with crates.io dependencies very soon (the `acvm` PR has already been merged and I believe a release is coming, and `ark-circom` also has supported arkworks 0.5.0). However, it is still very nice to have fixed revs now, so we no longer need to worry about upstream changes that accidentally break sonobe!

PR #209, opened 2025-02-18
Lift state length to the type system level

pmikolajczyk41:
Similarly to what happened with `external_inputs` fields in https://github.com/privacy-scaling-explorations/sonobe/pull/191 we move `state_len` to type parameter list.

### DIFF
--- a/examples/noname_full_flow.rs
+++ b/examples/noname_full_flow.rs
@@ -58,21 +58,23 @@ fn main() -> Result<(), Error> {
     ];
 
     // initialize the noname circuit
-    let f_circuit_params = (NONAME_CIRCUIT_EXTERNAL_INPUTS.to_owned(), 2); // state len = 2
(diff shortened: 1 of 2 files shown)


winderica (approved):
Looks good to me, thanks!

My only concern is that, with the state length as a const generic parameter, users who do not know the state length in advance may become unable to construct `FCircuit`s. But I think it is unlikely to happen in our case, because the user already has the circuit produced by alternative frontends at hand, and thus the state length is also known.

pmikolajczyk41:
> users who do not know the state length in advance may become unable to construct FCircuits

is this different to external input length? I mean, are there situations where you'd know external input structure, but not the IVC state?

winderica:
Ah thanks for pointing out that! I just realized that the external input length is already constant. It makes perfect sense to have state length as constant as well, since it is really weird if the user knows one but does not know another. And given a circuit compiled by external tools, we can be confident that both parameters are known.

PR #213, opened 2025-02-19
Opaque Solidity verifier

pmikolajczyk41:
# Simplified API for on-chain Nova proof verification  

## Summary  
This PR introduces an interface for **opaque verification** of Nova ZK proofs in Solidity. The goal is to abstract away low-level details of the proving system, enabling users to verify proofs without manually passing a dozen cryptographic elements.  

## Motivation  
Currently, any contract that wants to verify Nova proofs using Sonobe's Solidity verifier must explicitly pass numerous proof elements, each with deep cryptographic meaning. This makes usage cumbersome and error-prone, as seen in [[ChallengeHub.sol](https://github.com/zero-savvy/vimz/blob/master/on-chain-verification-showcase/ChallengeHub.sol)](https://github.com/zero-savvy/vimz/blob/master/on-chain-verification-showcase/ChallengeHub.sol), where the verification call requires declaring and passing multiple enigmatic things in a strict order.  

To address this, the new interface allows proof elements to be aggregated into a **single array**, with an option to provide the **number of steps**, **initial state**, and **final state** separately. These additional parameters may be useful for contracts that need to track or reuse them as public inputs.  

## Changes Introduced  
- **New interface for opaque proof verification**  
  - Accepts proof as a single array instead of multiple arguments  
  - Accepts `steps`, `initial_state`, `final_state` and the rest of proof data as a single array  
- **Improved usability**  
  - Reduces contract call complexity  
  - Makes verification logic more readable and maintainable  
  - Facilitates integration with higher-level protocols  
- **Backward compatibility**  
  - The existing low-level API remains unchanged for those who need fine-grained control  

### **Before (Using Low-Level API in ChallengeHub.sol)**  
```solidity
bool success = verifier.verifyNovaProof(
    i_z0_zi, 
    U_i_cmW_U_i_cmE, 
    u_i_cmW, 
    cmT_r, 
    pA, 
    pB, 
    pC, 
    challenge_W_challenge_E_kzg_evals, 
    kzg_proof
);
```  

### **After (Using New Opaque API)**  
```solidity
bool success = verifier.verifyOpaqueNovaProof(proof);
// or
bool success = verifier.verifyOpaqueNovaProofWithInputs(i, z0, zi, proof);
```  

Looking forward to your feedback! 🚀


solidity-verifiers/src/evm.rs
```diff
@@ -71,10 +71,27 @@ pub fn compile_solidity(solidity: impl AsRef<[u8]>, contract_name: &str) -> Vec<
...
 /// hence, we need to find the correct binary.
 fn find_binary(stdout: &str, contract_name: &str) -> Option<Vec<u8>> {
-    let start_contract = stdout.find(contract_name)?;
-    let stdout_contract = &stdout[start_contract..];
-    let start = stdout_contract.find("Binary:")? + 8;
-    Some(hex::decode(&stdout_contract[start..stdout_contract.len() - 1]).unwrap())
+    let intro_str = format!("======= <stdin>:{contract_name} =======\nBinary:\n");
+
+    println!("intro_str: {:?}", intro_str);
+    println!("stdout: {:?}", stdout);
```
winderica, line 77: Could we have these `println` calls disabled in release mode? It is okay to have these information when debugging, but end users may find it confusing when using the library.

Ideally, maybe we can have a feature (for example, `print-trace` just as `ark-std/print-trace`) that controls all debug information include this. But simply hiding `println`s behind something like `#[cfg(debug_assertions)]` is also fine with me.

solidity-verifiers/templates/nova_cyclefold_decider.askama.sol
```diff
@@ -144,4 +168,63 @@ contract NovaDecider is Groth16Verifier, KZG10Verifier {
 
         return(true);
     }
+
+    /**
+     * @notice  Verifies a Nova+CycleFold proof given initial and final IVC states, number of steps and the rest proof inputs concatenated.
+     * @dev     Simply reorganization of arguments and call to the `verifyNovaProof` function.
+     */
+    function verifyOpaqueNovaProofWithInputs(
```
winderica, line 176: Could you add unit tests for `verifyOpaqueNovaProofWithInputs` and `verifyOpaqueNovaProof`, in a similar way to `nova_cyclefold::tests::nova_cyclefold_solidity_verifier`?

By the way, I ran these functions locally with `revm` and they worked great, and as a side question, do they also work on onchain EVMs? I am not sure if onchain EVMs have any stricter restrictions than `revm` on parameter length/stack size/etc.

winderica (changes requested):
Thanks for proposing the Opaque APIs for solidity verifiers! I agree with everything you said, and the new APIs are indeed more succinct & error-prone.

I only have a few suggestions, while everything else looks good to me!

solidity-verifiers/src/evm.rs
pmikolajczyk41: my best apologies - these were just leftovers from my debugging... I've removed them

solidity-verifiers/templates/nova_cyclefold_decider.askama.sol
pmikolajczyk41: sure, sorry for skipping that initially - I wanted to firstly check if the change is okay with you, before writing tests :sweat_smile: 

now, we run old test scenarios with all 3 methods

pmikolajczyk41: > do they also work on onchain EVMs

yes! I have deployed the contract to Aleph Zero EVM Testnet and called each of the function successfully - please check:

https://evm-explorer-testnet.alephzero.org/address/0xD96A7C3D1A45D6A01Ac2A50097FB1D95F9349Adf?tab=txs

winderica: > > do they also work on onchain EVMs
> 
> yes! I have deployed the contract to Aleph Zero EVM Testnet and called each of the function successfully - please check:
> 
> https://evm-explorer-testnet.alephzero.org/address/0xD96A7C3D1A45D6A01Ac2A50097FB1D95F9349Adf?tab=txs

That's awesome! A huge thanks for experimenting with the new contract APIs. It is really nice to know everything works perfect onchain!

PR #203, opened 2025-02-06
perf: optimize pedersen and ipa commit

yelhousni:
# Description
This PR optimizes Pedersen and IPA commitments gadgets, through the use of joint_scalar_mul_be and scalar_mul_le implemented here https://github.com/arkworks-rs/r1cs-std/pull/155.
solves https://github.com/privacy-scaling-explorations/sonobe/pull/201#issuecomment-2629296814.
# Benchmark
The Pedersen check circuit goes from 3,575,519 r1cs to 2,455,364 r1cs.
The IPA verify circuit also goes from 77,175 r1cs to 73,367 r1cs.

PR #227, opened 2025-07-05
Fix performance regression in Pedersen circuit

### LINKED ISSUE #211: Investigate the performance regression in onchain decider
The onchain decider SNARK without the `light-test` feature now takes longer time and more RAM. In fact, the setup and proof generation could be done with 32GB of RAM in 2 minutes, as reported in https://github.com/privacy-scaling-explorations/sonobe/pull/88. However, the onchain decider now takes more than 3 minutes and eventually fails to run on my machine with 64GB of RAM:

```
> cargo test --package folding-schemes --lib --release -- folding::nova::decider_eth::tests::test_decider --exact --nocapture
    Finished `release` profile [optimized] target(s) in 0.19s
     Running unittests src/lib.rs (target/release/deps/folding_schemes-850178000889a0b5)

running 1 test
Start:   KZG10::Setup with degree 65536
··Start:   Generating powers of G
····End:     Generating powers of G ................................................87.030ms
··Start:   Generating powers of gamma * G
····End:     Generating powers of gamma * G ........................................91.847ms
··Start:   Generating negative powers of h in G2
····End:     Generating negative powers of h in G2 .................................296ns
··End:     KZG10::Setup with degree 65536 ..........................................182.324ms
Start:   Groth16::Generator
··Start:   Constraint synthesis
····End:     Constraint synthesis ..................................................17.176s
··Start:   Inlining LCs
test folding::nova::decider_eth::tests::test_decider has been running for over 60 seconds
error: test failed, to rerun pass `-p folding-schemes --lib`

Caused by:
  process didn't exit successfully: `/home/winderica/ivc/folding-schemes/target/release/deps/folding_schemes-850178000889a0b5 'folding::nova::decider_eth::tests::test_decider' --exact --nocapture` (signal: 9, SIGKILL: kill)
```

winderica:
This PR fixes #211 by fixing the `SNARKVerificationFail` error and addressing the high RAM usage.

Specifically, for the former issue, the problem lies in inputs of odd length `n`. In this case, the circuit first processes the `n - 1`-th element, and then in the `step_by` loop, `i` is in `1, 3, ..., < n - 2`, meaning that the loop body will access elements at indexes `0`, `1`, `2`, `3`, ..., `n - 5`, and `n - 4`. We can see that the `n - 2`-th and the `n - 3`-th elements are ignored. The fix simply replaces the incorrect range with the correct one `1, 3, ..., < n - 1`.

The second issue is because, while #203 reduces the size of Pedersen circuit, it introduces some linear combinations related to the `y` coordinate of the point, which are expensive to process. https://github.com/winderica/r1cs-std/commit/454466ba8801d42a73d0772a596363bcb7d92349 fixes this by avoiding the creation of some LCs. Now, the Nova decider runs in ~70s on my machine, which is 1.25x faster than the original decider in #88.


pmikolajczyk41:
that's awesome that you found it! at some point I've tried debugging this, but ended up without clear solution... thank you very much!

out of curiosity - are you planning to work more back in the Sonobe project, or was it rather a required bug fix?

winderica:
Hey @pmikolajczyk41, thank you for your interest in the status of Sonobe!

On one hand, the maintainers of the project have been pretty busy in other projects, which is why the development of the library has been stale for months.

On the other hand, personally I am still super enthusiastic about folding, and I want to add some ambitious features such as GPU acceleration, lookup arguments, faster decider, etc. to the project.

The basic implementation for them has already been done for a long while. However, to integrate it to the main repo while maintaining high code quality, I realized a refactor of Sonobe is necessary to make things more generic and modular, so we won't need to do the same implementation of these features for all supported folding schemes.

I have been working on this locally whenever I have time, but I will push the changes more actively so the community can learn the status of development. Although I cannot offer any guarantees, I will try to finish the refactor and features as soon™️ as possible 😁

PR #235, opened 2025-10-29
fix: fix serialization mismatch on wasm32

kbizikav:
I noticed that the IVC proof generated for the wasm32 target fails to be verified by the Rust native IVC proof verifier.

The cause was that `ArithSerializer` produces different results depending on the architecture’s bit width when calling `to_le_bytes()` on usize dimensions.


winderica (approved):
Thank you for spotting this! While this PR is targeting `main`, I will also pay attention to the serialization of `usize` when refactoring stuff on the `dev` branch.

